PR #49, opened 2014-06-24
Support custom sharding (not just logstash)

kaecyra:
0.3.1 allows index sharding by date, but the method is named for logstash which seems incorrect and restrictive. This PR generalizes the code but still allows quick logstash configuration in the same way as before, while leaving the door open for other pre-configurations as well (flume?).

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,13 @@
 ## Changelog
 
-### Future
+### 0.4.0
+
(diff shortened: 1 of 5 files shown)


dterror:
Thanks @kaecyra, sorry it took a while to feedback. A lot of things changed in the code to fill some of gaps you were addressing. As things stand now, I'm thinking this adds more code than value. But if you're still very interested in it, we can get re-open the discussion (maybe as an issue).

PR #82, opened 2014-11-29
add msec_key to have msec in @timestamp

shivaken:
If the plugin send time with format "%Y-%m-%dT%H:%M:%S.%L%z" which include milli second,
Elasticsearch is able to store the time with milli second.

But normally, 'time' passed by fluentd doesn't include milli second information. So, milli second information should be include as a separated value.

This patch enables fluentd-plugin-elasticsearch to make a 'timestamp' value from 'time' and the separated milli second value.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -124,9 +124,9 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
     chunk.msgpack_each do |tag, time, record|
       next unless record.is_a? Hash
       if @logstash_format
-        if record[@msec_key]
+        if record.has_key?("@msec_key")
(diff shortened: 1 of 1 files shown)


salekseev:
+1 for https://github.com/slaunay/fluent-plugin-elasticsearch/commit/4423a001bfe8d36356cafca3b7657262459acc3c

dterror:
@shivaken are you still interested in this? Things have changed, so we don't need `msec_key` since we now have `@time_key`. We'd only need to format the resulting `@timestmap` string to include millisecond.

shivaken:
Hi @dterror, For me, this pull request is not so important already as you said. Only one motivation was for beginner or for guys who don't know how to implement millisec. I thought if that requires a lot of effort, that is not so good. Therefore I just proposed this, but I understand you feel it's feature creep and agreed.
So, maybe it's better to implement another formatting plugin.
Please feel free to reject this PR. Thanks.

dterror:
thanks @shivaken

PR #86, opened 2015-01-07
Convert binary strings to utf-8 before creating json

### LINKED ISSUE #77: Does fluent-plugin-elasticsearch convert to UTF-8 before creating json?
I have an error with sending some data to elasitcsearch using this plugin, some examples of the error:

Caused by: org.elasticsearch.common.jackson.core.JsonParseException: Invalid UTF-8 start byte 0xa3
Caused by: org.elasticsearch.common.jackson.core.JsonParseException: Invalid UTF-8 middle byte 0x6d
Caused by: org.elasticsearch.common.jackson.core.JsonParseException: Invalid UTF-8 middle byte 0x20

From the research I have done ( http://stackoverflow.com/questions/13830346/jackson-json-parser-invalid-utf-8-start-byte ; http://stackoverflow.com/questions/24629013/jmeter-invalid-utf-8-middle-byte ) it could be that the json being sent by the plugin is not UTF-8  - and json should always bt URD-8 which I am sure you know.

Could you confirm that fluent-plugin-elasticsearch converts to UTF-8 before crafting the json request?

Thanks.

salekseev:
This is to prevent the following errors in case of high bit characters in log messages:
Caused by: org.elasticsearch.common.jackson.core.JsonParseException: Invalid UTF-8 start byte 0xff

Also should fix #77.

### DIFF
--- a/fluent-plugin-elasticsearch.gemspec
+++ b/fluent-plugin-elasticsearch.gemspec
@@ -19,6 +19,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'fluentd', '>= 0.10.43'
   s.add_runtime_dependency 'patron', '~> 0'
   s.add_runtime_dependency 'elasticsearch', '>= 0'
+  s.add_runtime_dependency "yajl-ruby", "~> 1.0"
 
   s.add_development_dependency 'rake', '~> 0'
   s.add_development_dependency 'webmock', '~> 1'
--- a/fluent-plugin-elasticsearch.gemspec
+++ b/fluent-plugin-elasticsearch.gemspec
@@ -3,7 +3,7 @@ $:.push File.expand_path('../lib', __FILE__)
 
 Gem::Specification.new do |s|
   s.name          = 'fluent-plugin-elasticsearch'
-  s.version       = '0.7.0'
+  s.version       = '0.7.1'
   s.authors       = ['diogo', 'pitr']
   s.email         = ['pitr@uken.com', 'diogo@uken.com']
   s.description   = %q{ElasticSearch output plugin for Fluent event collector}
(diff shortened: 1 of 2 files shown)


pitr:
JSON only works with unicode strings (http://www.json.org/)
Json string spec says UNICODE

salekseev:
Exactly, but it's quite possible to feed binary data into fluentd via MessagePack (our case) or other means which blows up upon writing to ElasticSearch as it only supports JSON and hence utf-8 charset.

pitr:
I think you should take care not to send non-utf8 data to this plugin

dterror:
+1 for utf8 only handling

salekseev:
It it certainly an option to have a separate plugin do the non-utf8 chars sanitation, my thought was that since this plugin's only job is to send to elasticsearch (which only supports utf-8), it should perform the sanitation in case it receives binary data from fluentd (which it's absolutely able to send).

In our particular use case I prefer not to mess around with tags as we also are taking advantage of  https://github.com/tagomoris/fluent-plugin-forest that relies on tags.

If you have a strong opinion about not including this functionality, would you consider it if it was enabled by an option?

salekseev:
Seems I have achieved the same effect by combining forest and record_modifier plugins like:
```
  <match production.*>
    subtype record_modifier
    type forest
    <template>
      char_encoding ISO-8859-1:utf-8
      tag utf8_production.${tag_parts[1]}
    </template>
  </match>
  <match utf8_production.*>
    subtype elasticsearch
    type forest
    <template>
      buffer_chunk_limit 32m
      buffer_path /var/log/fluentd/buffer/production.${tag_parts[1]}
      buffer_queue_limit 512
      buffer_type file
      disable_retry_limit true
      flush_interval 3s
      hosts 127.0.0.1:9200
      logstash_dateformat %Y.%m.%d
      logstash_format true
      logstash_prefix production
      max_retry_wait 600s
      num_threads 2
      retry_wait 5s
      type_name ${tag_parts[1]}
      utc_index true
    </template>
  </match>
```

PR #89, opened 2015-01-30
Add support for arbitrary operation, instead of simply index.

jmoseley:


dterror:
@jmoseley we're trying to restrict feature creep.Maybe  if you share your use case, it might fire up the discussion. Also in the case we decide to go with it, we'd probably have to be more careful with failures, since we don't want to be blindly re-trying legit create failures.

jmoseley:
@dterror Sure thing. In my case, each of my records has a unique ID and we are using this ID for the ES storage of the record. Occasionally we have records that come through the system multiple times to ensure that all logs were consumed (since Fluentd is an 'at most once' system). In these cases the elasticsearch performance would be increased if it could simply ignore existing records (as the 'create' command does) instead of indexing again (the way the 'index' command does).

There might also be use to remove records, but I have no need of that functionality.

dterror:
I see. thks man.

As I said, we'd have to change the `send` logic to avoid hitting ES with unnecessary retries. The benefit of `create` over `index` would be low unless there's a great volume duplicates. Those operations seem to make more sense in environments in which you have more granular control over your records, like full-text-indexing database records. For this plugin, for instance, I can't see `delete` ever working.

I'm tending towards -1

@pitr inputs?

jmoseley:
With that in mind, could we constrain this to be either index or create? I believe the failure cases are similar, and should work.

pitr:
Perhaps we can move "create" functionality in a separate plugin, sort of like how fluentd-plugin-mongo has multiple output plugins.

However, I am unsure what performance advantage there is in using "index" over "create". Both need to *index* new records. Former just discards duplicated, which should not be expensive. Is there a performance comparison of those operations somewhere? Why do you believe "create" is faster?

jmoseley:
From my understanding the create operation will compare the ids and throw away duplicates before any indexing happens [1]. The index operation will index the new document and overwrite the existing one. The plugin would need to be able to handle these failures and ignore them, but I believe the performance is faster.

[1] http://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#operation-type

pitr:
If "create" is indeed faster than "index", I believe that should be the default behaviour when `id_key` or `parent_key` are provided. However, we'd need hard numbers :stuck_out_tongue:

pitr:
fixed in v1.3.0 by #143

jmoseley:
:+1:

PR #91, opened 2015-03-10
Dynamic index based on tag parts (split on ".")

jimmidyson:


repeatedly:
Using forest plugin is not fit for your case?

https://github.com/tagomoris/fluent-plugin-forest

jimmidyson:
@repeatedly - it says in forest plugin that "You SHOULD NOT use ForestOutput for tags increasing infinitely" & that's the problem in my case: some of the parts of the tags I want to use are generated so tags could increase infinitely... This way is pretty simple but works well for my use case.

dterror:
I understand the appeal of using tag_parts, but I feel like there should be a general solution. The next thing that's going to happen is people wanting to use `tag_parts` for the `index` name or the `_type` and then people wanting to use `escaped_tag` or `hostname` similar to what forrest nicely provides.

So I think those are general enough to motivate another meta-plugin, such as forrest, but circumventing its limitations. I know it might be a big challenge, but it's better to tackle it in that level of abstraction rather than creeping concerns in plugin-elasticsearch.

What do you think @pitr?

jimmidyson:
@dterror @pitr I agree a general solution would be great, but this is something that is necessary for many use cases - see the other 2 issues. The `tag_parts` strategy is already used in a number of other fluentd plugins (e.g. https://github.com/sonots/fluent-plugin-record-reformer, https://github.com/bwalex/fluent-plugin-docker-format).

How about we put it in for now & migrate to a more general solution when one is available?

pitr:
There's a lot of cost associated with adding new features, as we will need to support them indefinitely. Can't just remove configuration in the future. A lot of people simply install latest version of fluent-plugin-elasticsearch and if their configuration worked for them before, it SHOULD keep working.

I agree with @dterror's stance. Output plugins (like this one) should not be dealing with complicated routing. Perhaps it's time to restart the discussion in the forest plugin to address their limitation tagomoris/fluent-plugin-forest#15 That will benefit a LOT of plugins.

PR #92, opened 2015-03-17
Switch to using curb elasticsearch http transport instead of patron

salekseev:
This will hopefully make things faster (needs benchmarking) and will reduce compatibility issues with newer libcurl and parton gems.


salekseev:
The tests are failing and need fixing. I'm trying to figure out why.

pitr:
I wish we could move away from libcurl-based games altogether

dterror:
Agree with @pitr

salekseev:
I need to do some benchmarking to see how it compares in performance / mem usage to faraday+patron vs curb vs faraday+excon (for example for pure ruby).

salekseev:
Closed in favor of #93.

PR #93, opened 2015-03-18
Switching to using excon faraday adapter to remove libcurl dependency.

salekseev:
This is another option. Passes tests but requires performance benchmarking prior to merging. I'll let you know what I find in terms of performance as soon as I'm able to put it under load.

### DIFF
--- a/fluent-plugin-elasticsearch.gemspec
+++ b/fluent-plugin-elasticsearch.gemspec
@@ -3,9 +3,9 @@ $:.push File.expand_path('../lib', __FILE__)
 
 Gem::Specification.new do |s|
   s.name          = 'fluent-plugin-elasticsearch'
-  s.version       = '0.7.1'
-  s.authors       = ['diogo', 'pitr', 'salekseev']
-  s.email         = ['pitr@uken.com', 'diogo@uken.com', 'stas.alekseev@gmail.com']
+  s.version       = '0.7.0'
+  s.authors       = ['diogo', 'pitr']
+  s.email         = ['pitr@uken.com', 'diogo@uken.com']
   s.description   = %q{ElasticSearch output plugin for Fluent event collector}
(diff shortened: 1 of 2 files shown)


repeatedly:
Is there a concern for merging this PR? Performance?
If someone has a dataset, I also can check the performance.

salekseev:
We have been running this version in our development environment for about a month now with no apparent issues so it has passed a smoke test. We don't nearly have the throughput in dev that matches our production and I haven't gotten around to making a synthetic benchmark for it yet.

dterror:
Thanks @salekseev, I'll run a simple benchmark, but in principle removing libcurl dep will be for the best.
Can you remove your version and authorship changes to gemspec? I'll merge more PRs before bumping.

repeatedly:
> removing libcurl dep will be for the best.

I agree and elasticsearch plugin doesn't flush events in sub-second interval.
So small performance degression is not huge problem.
If need more performance, then we can consider other library like httpclient.

http://reevoo.github.io/blog/2014/09/12/http-shooting-party/

salekseev:
@dterror, we should be all set with gemspec changes.

@repeatedly, I chose excon based on its feature set and popularity (aliveness) after reading http://www.slideshare.net/HiroshiNakamura/rubyhttp-clients-comparison.

dterror:
thks @salekseev, I'll release a new version this week.

repeatedly:
@salekseev nahi-san, a.k.a author of httpclient gem and this presentation, is my co-woker. So if we hit the problem with excon in the future, we can ask him for choosing other HTTP client :)

@dterror Good! I can't wait new version :+1:

dterror:
@repeatedly new version is out v0.8.0

repeatedly:
@dterror Thanks!

@salekseev Awesome work :+1:

PR #96, opened 2015-03-29
Allow `tag` to be used as `_type` property.

izaakschroeder:
Take the last tag component and set it as the `_type` property. e.g. `foo.bar.baz` gets `_type: "baz"`. For something like the following use case:

```javascript
fluent.configure('foo.bar');
fluent.emit('baz', { ... });
```

It is quite handy to capture the `baz` as the `_type` property. I think this is a common enough behavior to warrant a configuration option for. It's also impossible to configure with any of the existing options; using `tag_name "#{tag}"` fails since the evaluation happens when the config file is parsed, not when the message is processed.

Specs added for this new behavior.


dterror:
Thanks @izaakschroeder. I see the problem pattern this is trying to solve and we're discussing more general solutions in #91.

PR #98, opened 2015-04-14
Switching to httpclient faraday transport

salekseev:
People at #93 seemed to have a preference for httpclient over using excon. I have tested out this configuration in dev and it passes the smoke test. No benchmarking has been done.


dterror:
I've already merged #93 @salekseev, things are good, if we have a problem in the future, we'll pick another http client.

PR #106, opened 2015-05-21
Add in allow_overrides feature

gtrevg:
A feature that allows other fluentd collectors to determine either the `_index`, `_id`, `_type`, or `_parent` values before pushing the data to the elastic search fluentd collector.

### DIFF
--- a/History.md
+++ b/History.md
@@ -2,6 +2,9 @@
 
 ### Future
 
+### 0.9.0
+- Add new `allow_overrides` option
+
 ### 0.8.0
 - Replace Patron with Excon HTTP client
 
--- a/README.md
+++ b/README.md
@@ -202,6 +202,28 @@ Please consider using [fluent-plugin-forest](https://github.com/tagomoris/fluent
 </match>
 ```
 
+---
+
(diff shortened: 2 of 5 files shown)


pitr:
Thanks for PR, but this seems overly magical and unnecessary. It appears that you are mixing logs. Why not use a separate tag?

```
<match app.logs>
  type elasticsearch
  index_name _index
</match>
<match www.logs>
  type elasticsearch
  index_name _index
  id_key _id
</match>
```

Alternatively, maybe https://github.com/sonots/fluent-plugin-record-reformer can help.

gtrevg:
The reason was because of ownership or servers.  The management of a bank of fluentd collectors and the elastic search cluster that they are tied to are owned by a team that wouldn't be updating their fluentd configs every time a new project came online and needed to push data to a different index.

This would allow Cluster A of fluentd machines to massage log data specific to the application they are managing.  Then that data enters Cluster B via the standard fluentd forwarding protocal and not via direct access to elastic search.  The target index could be set by Cluster A and all the elastic search management, passwords, heartbeats, & etc would be managed within Cluster B.

thanks

PR #116, opened 2015-07-09
Add supports for dynamic index & type names. add `remove_keys` feature

cherrot:
Although I can do this with [fluent-plugin-forest](https://github.com/tagomoris/fluent-plugin-forest), but it would be memory
consuming if there were tons of different tags of fluentd events since
`fluent-plugin-forest` creates new plugin instance for each template.
So I implement this inside this plugin.
You can use the following configuration to gain this feature:

    <match fluent.*>
      type null # to prevent recursively logging
    </match>
    <match **>
         type elasticsearch
         logstash_format true
         logstash_prefix ${my_index_key}
         type_name ${tag}
         remove_keys my_index_key,foo
     </match>

Suppose you send a message with tag `test.es`:
`{"foo":"u won't see me", "my_index_key": "myfluent", "bar": "Hello from fluentd"}`

You will get the following document from the index `myfluent-YYYY.MM.DD` of elasticsearch:
`{"bar": "Hello from fluentd"}`

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -28,7 +28,8 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :reload_connections, :bool, :default => true
   config_param :reload_on_failure, :bool, :default => false
   config_param :time_key, :string, :default => nil
-  config_param :ssl_verify , :bool, :default => true
+  config_param :ssl_verify, :bool, :default => true
+  config_param :remove_keys, :string, :default => nil
 
   include Fluent::SetTagKeyMixin
   config_set_default :include_tag_key, false
@@ -39,6 +40,9 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
 
   def configure(conf)
     super
+    if @remove_keys
+      @remove_keys = @remove_keys.split(',')
+    end
   end
 
   def start
@@ -48,16 +52,17 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   def client
     @_es ||= begin
       adapter_conf = lambda {|f| f.adapter :excon }
-      transport = Elasticsearch::Transport::Transport::HTTP::Faraday.new(get_connection_options.merge(
-                                                                          options: {
-                                                                            reload_connections: @reload_connections,
-                                                                            reload_on_failure: @reload_on_failure,
-                                                                            retry_on_failure: 5,
-                                                                            transport_options: {
-                                                                              request: { timeout: @request_timeout },
-                                                                              ssl: { verify: @ssl_verify }
-                                                                            }
-                                                                          }), &adapter_conf)
+      transport = Elasticsearch::Transport::Transport::HTTP::Faraday.new(
+        get_connection_options.merge(
+          options: {
+            reload_connections: @reload_connections,
+            reload_on_failure: @reload_on_failure,
+            retry_on_failure: 5,
+            transport_options: {
+              request: { timeout: @request_timeout },
+              ssl: { verify: @ssl_verify }
+            }
+          }), &adapter_conf)
       es = Elasticsearch::Client.new transport: transport
 
       begin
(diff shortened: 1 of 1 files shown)


jimmidyson:
@cherrot :+1: I had a similar, although less feature rich, PR in to use dynamic index & it was rejected. I hope this one gets accepted.

jeesim2:
'_type name by key' is essential feature to me.
Because ES create type mapping to each _types, various data in same _type can cause type conflicts problem.
this PR will make it easier to seperate various data to each _types.

dterror:
Hey @repeatedly @tagomoris, any plans on Fluent side to support this type of interpolation as part of the plugin API. This seems to be the most frequent topic of discussion around this plugin and I imagine a lot of others as well.
c.c @pitr

jeesim2:
@dterror any issue to have this feature in fluent-elasticsearch-plugin?
or do you think it's better to provide this feature by fluent core?
I have no idea what are you exactly asking to @repeatedly.
thanks :)

dterror:
@jeesim2 mostly following up the discussion on https://github.com/uken/fluent-plugin-elasticsearch/pull/91

cherrot:
@dterror  I have considered to add this feature of dynamically configuring fluentd plugin by using expressions like `${blabla}`,  and found it's not easy to accomplish this based on the current plugin mechanism.  For now I suggest implementing these at per-plugin level and waiting for  `fluentd`'s updates :)

I'm not an expert on ruby anyway, so @jeesim2  is there any plan for supporting dynamic configuration in fluent core?

jeesim2:
@cherrot I agree with your idea.
if this PR have no critical problem, It would be better that grant this PR and await Fluent-core level support enhancement.
I'm sorry but I'm not a ruby guy too. : (

repeatedly:
In fluentd core, we have a plan to support similar feature in Plugin API but we don't have enough experience.
If fluent-plugin-elasticsearch supports this feature, it is a good as a basis for discussion.

cherrot:
Thanks for paying attention to this.
When I say

> it's not easy to accomplish this based on the current plugin mechanism

I'm saying that at first I was considering to make `fluentd` "re-initialize" the config params when an event comes (and has not passed to the underlying plugin yet). And I found for now `fluentd` does not have such a "pre-hook" to handle this. 
I'll appreciate it if someday `fluentd` supports this feature ;) Furthermore, what the param `remove_keys` does could move into a common "post-hook" before invoking `write`, right?

repeatedly:
BTW, you can use `record_transformer` filter to remove keys.

http://docs.fluentd.org/articles/filter_record_transformer#removekeys-optional-string-type

dterror:
That's sweet, thanks @repeatedly

cherrot:
But what if I need this key as a dynamic config param and then delete it before writing to es?  ;-) 
Maybe this is not a essential feature, feel free to remove it (or tell me to push a revert commit)

ewolinetz:
@cherrot this is great!  I was wondering if it would be possible to allow this to grab the value of a nested hash as well?  e.g. record => { second_hash => { nested: value} }  
Currently if i try something like ${second_hash.nested} it isn't able to find pull the value out...

cherrot:
@ewolinetz Hmm, that's a good question... Currently nested json value is not supported. And I doubt whether it is necessary? For me, I prefer to keep it simple ;)
Related discussion is on #122

ewolinetz:
@cherrot completely understand!  It looks like I should be able to get this functionality using your changes with https://github.com/kazegusuri/fluent-plugin-flatten-hash

cherrot:
Wow that's a great mashup! Thanks @ewolinetz 
And... @dterror  sorry for bothering, do we need any further discussions on this PR?

dterror:
No worries. Let's concentrate the discussion on https://github.com/uken/fluent-plugin-elasticsearch/issues/122

pitr:
Thank you @cherrot for PR. I'm going to close this for now, since @dterror and I don't think that this is the way to go with supporting dynamic configurations. But this and @jimmidyson PR (and everyone involved in discussion here) show us that this is a feature that is very much needed.

@dterror opened an issue #122 where we can keep talking about a solution that works for everyone.

cherrot:
@pitr :+1:   Sorry for the late reply. I'll keep track of #122 and #127 
The new feature design looks great, thanks :)

PR #119, opened 2015-07-17
Add support for excess of log size

nakc:
When the size of fluent log exceeds 32KB, Elasticsearch returns error.
However this plugin sends this exceeded log for anytime.
So I modified so that send only log that less than 32KB(@max_size).

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -3,6 +3,7 @@ require 'date'
 require 'excon'
 require 'elasticsearch'
 require 'uri'
+require 'objspace'
 
 class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   class ConnectionFailure < StandardError; end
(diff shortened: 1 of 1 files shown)


dterror:
32KB is the maximum _term_ size in Lucene, it's not necessarily a limitation on the _record_ size. Because the definition of what a 'term' is is highly dependent on your record semantics and how you create its mapping in elasticsearch, enforcing any sort of limitation is out of the scope (and competence) of this plugin.

What's likely happening is that ES is storing your whole record as one big term, probably using a field with `'index':'not_analyzed'`. I'd recommend you check out this doc on [core types](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-core-types.html).

yosssi:
@dterror How about adding the option which checks each `String` value's size of the `record` object instead of the whole `record` size? I need this option for my project to prevent Fluentd from trying to send log data having large term data to Elasticsearch many times because that impacts the machine resources of both Fluentd and Elasticsearch server.

dterror:
Even if we enforce size of each field, it might be that you have ES configured to interpret the whole record as one big term. Similarly, you could have each field configured with an analyzer that tokenizes its contents into multiple smaller terms.

PR #120, opened 2015-07-25
Add secret parameters

cosmo0920:
Fluentd's config_params now supports concealing given parameters with xxxxxx as below:

```log
2015-07-25 16:48:57 +0900 [info]: using configuration file: <ROOT>
  <match my.logs>
    host localhost
    user fluent
    password xxxxxx
    type elasticsearch
    include_tag_key true
    tag_key _key
  </match>
</ROOT>
```

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -12,7 +12,7 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :host, :string,  :default => 'localhost'
   config_param :port, :integer, :default => 9200
   config_param :user, :string, :default => nil
-  config_param :password, :string, :default => nil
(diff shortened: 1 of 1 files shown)


dterror:
Thanks @cosmo0920

cosmo0920:
Thanks! :yum:

pitr:
Hmm, wouldn't this will depend on fluentd v0.12.13 (currently we have `>= 0.10.43` in our gemspec) - https://github.com/fluent/fluentd/blob/4e3a4ecb/ChangeLog#L34

dterror:
It won't break if you have < 0.12.13, it just gets ignored. But maybe we should make it clear in the documentation.

cosmo0920:
Yes, it just get ignored when you use fluentd < 0.12.13.

pitr:
Thanks for clarifying. WOW you guys are fast! Took under 10 minutes!

PR #123, opened 2015-08-10
Updating out_elasticsearch.rb to pass client configurations to Excon …

ewolinetz:
…for Client verification and specify a CA for host verification. Updated README with configuration examples.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -29,6 +29,10 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :reload_on_failure, :bool, :default => false
   config_param :time_key, :string, :default => nil
   config_param :ssl_verify , :bool, :default => true
+  config_param :client_key, :string, :default => nil
+  config_param :client_cert, :string, :default => nil
+  config_param :client_key_pass, :string, :default => nil
+  config_param :ca_file, :string, :default => nil
 
   include Fluent::SetTagKeyMixin
   config_set_default :include_tag_key, false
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -47,7 +51,8 @@ def start
 
   def client
     @_es ||= begin
-      adapter_conf = lambda {|f| f.adapter :excon }
+      excon_options = { client_key: @client_key, client_cert: @client_cert, client_key_pass: @client_key_pass }
+      adapter_conf = lambda {|f| f.adapter :excon, excon_options }
```
dterror: Since you're configuring Faraday's ssl options bellow (in the `transport_options` hash), do we need to configure Excon here as well?

ewolinetz: True, I could specify the client_key and client_cert in the Faraday SSL hash.  However it appears Faraday doesn't have a parameter for a client_key_password so people would need to use an unencrypted client key.

jimmidyson: @ewolinetz It's kind of normal for people to use keys without passwords in non-interactive apps sadly so using Faraday should cover most use cases IMO. But there will be others who do need a key password so...

ewolinetz: @jimmidyson The nice thing about this is you aren't required to use an encrypted key; only if you specify a value for client_key_pass will it try to decrypt the key.

pitr:
looks good to me, @dterror let's ship it?

dterror:
ya :sheep: it (sorry, I reviewed it and then totally forgot about it)

pitr:
thanks @ewolinetz!

Issue #122, opened 2015-08-05
Dynamic variables in config


dterror:
It seems like a lot of people want to use dynamic variables in configuration, mostly to retrieve index and type based on record and tag variables. The common resolution to this is using the forrest plugin, but it turns out there are a lot of problems with that plugin.

Ideally this plugin wouldn't have to deal with this concern, but as a lot of users seem to really need it, I think we should think of the best away to tackle it.

Related PRs/issues:
- https://github.com/uken/fluent-plugin-elasticsearch/pull/91
- https://github.com/uken/fluent-plugin-elasticsearch/pull/116
- https://github.com/fluent/fluentd/issues/596

@pitr

dterror:
Q: do we need to support record fields exposed as configuration objects?
For example:
```
<match **>
     type elasticsearch
     type_name ${a_key_in_my_object}
 </match>
```

And then at what depth would we allow it? Should we expose a jsonpath interface so that people refer to `${keys.in.arbitrary.levels}`?

pitr:
To me it seems that we essentially need to add a way for users to write code in their config. Otherwise, no amount of customisation will satisfy all the requirements.

PS. One thing I'd like to suggest is creating a separate plugin (e.g. `out_elasticsearch_dynamic.rb`) so we can experiment before impacting existing users of this plugin.

dterror:
+1 and +1

pitr:
How about allowing Ruby expanding using `${..}` syntax like [filter_record_transformer](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/filter_record_transformer.rb) but for most fields?

dterror:
I think we might have to do it in order to achieve full-flexibility. I think this is probably super unsafe and inefficient, but Fluentd is using it, so... idk.

pitr:
So would something like this work for everyone?

```
<match my.logs.*>
  type elasticsearch_unsafe
  host ${tag_parts[3]}.example.com
  port ${9200+rand(4)}
  logstash_prefix ${record['prefix'] || 'generic_prefix'}
  include_tag_key ${record['username'] != 'admin'}
  id_key ${record['my_id'] ? 'my_id' : 'id'}
</match>
```

ewolinetz:
@pitr +1

jimmidyson:
@pitr :+1:

jeesim2:
[...]

PR #126, opened 2015-09-02
Implement new param 'logstash_prefix_key'

ymattw:
This is useful when user wants to create a different index per app name.  See updated README.md for details.

Please let me know if there's a better way to implement this. Thanks!


dterror:
Because of the big demand for dynamic variables in the configuration, we're discussing a more general approach to solve that problem in https://github.com/uken/fluent-plugin-elasticsearch/issues/122

PR #127, opened 2015-09-02
initial changes to implement issue #122

ewolinetz:


ewolinetz:
@pitr @dterror PTAL here is what I have so far for #122 ..I see that the Travis tests failed, I'll investigate that further tomorrow.

dterror:
Cool @ewolinetz
One thing we were discussing is adding this feature to a new plugin type (like `elasticsearch_unsafe`, `elasticsearch_dynamic` or something), so it'd probably go in a different file and register as a different plugin. It seems like inheritance would be enough to share most of the core behaviour.

The reasoning is: Adding this level of dynamic evaluation is both extremely slow and unsafe and we don't want to penalize users that don't really need it (or can get around it in different ways).

ewolinetz:
it looks like travis ci isn't finding the test cases for out_elasticsearch_dynamic, but it is working locally when i run it.. @dterror I refactored out the dynamic changes into a separate file per your comment.

ewolinetz:
There currently seems to be an issue of the ES client trying to connect when port if something like ${9200 + rand(4)}.  It is able to evaluate it, however it doesn't seems to come back from `es.ping`.  If i specify port as ${9199 + 1} it does.  Even if I hack the code to only allow it to be randomly generated once, it doesn't...

ewolinetz:
@pitr @dterror I think this is at a pretty good state now.  
You can specify things like: 
  * hosts ${record['host1']}:9200,${record['host2']}:9200
  * index_name my_index.${Time.at(time).getutc.strftime(@logstash_dateformat)}
  * logstash_prefix ${tag_parts[3]}
  * port ${9200+rand(4)}
  * index_name ${tag_parts[2]}-${Time.at(time).getutc.strftime(@logstash_dateformat)}

PTAL

pitr:
You can add the new spec to .travis.yml here https://github.com/uken/fluent-plugin-elasticsearch/blob/master/.travis.yml#L9 Something like this

```
script:
  - bundle exec ruby -S -Itest test/plugin/test_out_elasticsearch.rb
  - bundle exec ruby -S -Itest test/plugin/test_out_elasticsearch_dynamic.rb
```

Also, can you add some basic info to README. Even just those examples in your last comment are fine.

Otherwise, +1 from me. @dterror? Since this doesn't impact anybody using the original plugin, let's get this out soon and see what people say.

dterror:
Alright, that sounds good as a plan.
+1 pending @pitr comments
Thanks @ewolinetz

ewolinetz:
@pitr @dterror both changes were made, do you want me to squash commits as well or are we set as is?

pitr:
Thanks, this is great. v1.1.0 has been released with this plugin

PR #133, opened 2015-10-24
Improve dynamic

repeatedly:
Two changes for elasticsearch_dynamic plugin

- `instance_variables` creates new array so prefetch these variables in `configure` method to avoid memory allocation.
- modifying `@dynamic_config` has race-condition when `num_threads` is larger than 2. Clone it to avoid this problem.


pitr:
Thanks @repeatedly!

pitr:
released in v1.2.0

PR #136, opened 2015-12-02
Added option "resurrect_after"

robertfirek:
Adding this option give possibility to avoid starvation of the pool from Elasticsearch Ruby client when requests number more frequent than the "resurrect_after" time.

+ Additional comment for `buffer_chunk_limit` option

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -27,6 +27,7 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :request_timeout, :time, :default => 5
   config_param :reload_connections, :bool, :default => true
   config_param :reload_on_failure, :bool, :default => false
+  config_param :resurrect_after, :time, :default => 5
   config_param :time_key, :string, :default => nil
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -27,6 +27,7 @@ class ConnectionFailure < StandardError; end
   config_param :request_timeout, :time, :default => 5
   config_param :reload_connections, :bool, :default => true
   config_param :reload_on_failure, :bool, :default => false
+  config_param :resurrect_after, :time, :default => 5
```
pitr: can you set default to 60, since it's current default - https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/transport/base.rb#L11

pitr:
Thanks for PR, makes sense to me. Could you add documentation for that config to README.md (and reference it in Table of Content too please) and address my other comment, and I can merge it

robertfirek:
@pitr I think it is ready. I don't how to introduce this change to dynamic version, so I left it unchanged.

pitr:
released in v1.2.0

I added it to dynamic version

PR #143, opened 2016-01-04
Support update operation

EthanLiou:
Hi Team,
I have a plan to use ES to store some user setting. I don't want to upload all settings to the backend once user changes just one setting. So I add a partial flag to control it.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -23,6 +23,7 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :type_name, :string, :default => "fluentd"
   config_param :index_name, :string, :default => "fluentd"
   config_param :id_key, :string, :default => nil
+  config_param :partial, :bool, :default => false
   config_param :parent_key, :string, :default => nil
   config_param :request_timeout, :time, :default => 5
   config_param :reload_connections, :bool, :default => true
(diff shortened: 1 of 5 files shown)


pitr:
Good looking PR, thanks @EthanLiou. My only concern is naming, perhaps `upsert` would be better than an ambiguous `partial`, what do you think? Also README should clarify that id field is required, since this code drops messages otherwise.

@dterror could you take a look?

EthanLiou:
Hi @pitr, I provided a general way to set write_operation. I think it's better than the previous one.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -154,17 +155,29 @@ def write(chunk)
...
       end
 
       if @parent_key && record[@parent_key]
-        meta['index']['_parent'] = record[@parent_key]
+        meta['_parent'] = record[@parent_key]
       end
 
-      bulk_message << meta
-      bulk_message << record
+      if @write_operation == "update" || @write_operation == "upsert"
```
pitr: `case .. when` would work better here

lib/fluent/plugin/out_elasticsearch_dynamic.rb
```diff
@@ -164,8 +164,20 @@ def write(chunk)
         host = "#{dynamic_conf['host']}:#{dynamic_conf['port']}"
       end
 
-      bulk_message[host] << meta
-      bulk_message[host] << record
+      if dynamic_conf["write_operation"] == "update" || dynamic_conf["write_operation"] == "upsert"
```
pitr: could you de-duplicate this code by extracting it into a function?

pitr:
This is even better, will close #89.

Thanks for supporting the dynamic version as well, usually i have to ask :)

EthanLiou:
Thanks for your suggestion:)

pitr:
Looks great!

@dterror is on vacation, I hope is having fun!

pitr:
v1.3.0 is released

dterror:
Thanks guys, this looks great!

Issue #149, opened 2016-01-25
bypass type_name


nnovikov:
If log file in JSON format have _type key, I want to bypass _type without change it to "fluentd" (by default).

pitr:
3 things I think you can do:

1. Use forest plugin and specify different values for _type key that you have. See https://github.com/uken/fluent-plugin-elasticsearch#not-seeing-a-config-you-need
2. Use `@type elasticsearch_dynamic`, see https://github.com/uken/fluent-plugin-elasticsearch#dynamic-configuration
3. Create a PR that implements the behaviour that you want. If you decided to do this, first let us know here how you are planning to implement it

pitr:
implemented in #169

PR #151, opened 2016-02-02
add bypass index and type

nnovikov:
I add some changes for bypass 'index' and 'type' fileds directly to elasticsearch. You need logstash_format=false in your config.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -165,15 +165,19 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
         else
           target_index = "#{@logstash_prefix}-#{Time.at(time).strftime("#{@logstash_dateformat}")}"
         end
+        target_type = @type_name
       else
-        target_index = @index_name
+	target_index = record.fetch("_index", @index_name)
+        target_type = record.fetch("_type", @type_name)
+        record.delete("_index")
+        record.delete("_type")
       end
 
       if @include_tag_key
         record.merge!(@tag_key => tag)
       end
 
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -165,15 +165,19 @@ def write(chunk)
         else
           target_index = "#{@logstash_prefix}-#{Time.at(time).strftime("#{@logstash_dateformat}")}"
         end
+        target_type = @type_name
       else
-        target_index = @index_name
+	target_index = record.fetch("_index", @index_name)
```
pitr: fix weird spacing

pitr:
I think this behaviour needs to be enabled by a toggle setting, off by default. That way existing users would not be impacted.

nnovikov:
I add this PR as example for this issue https://github.com/uken/fluent-plugin-elasticsearch/issues/149
If you approve this feature, I will add config parameter (what name you seem good?) and fix spacing. Also if you have any additional request, please tell me.

pitr:
Definitely approve. How about `index_name_from_record` with value signifying the key on record (eg. "_index"), default being `nil` in which case disabling this feature. And similarly, `type_name_from_record`.

Would be nice to have a test for each, and an entry in README

nnovikov:
I try to write test, but have error:
/usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- fluent/test (LoadError)
I need install fluentd? Or I can test module only?

pitr:
You need to run `bundle install` to get all dependencies

pitr:
implemented in #169

PR #153, opened 2016-02-04
Add support for `target_index_key` config.

ashb:
This is one of the PRs to come out of uken/fluent-plugin-elasticsearch#152 "Time.parse
is slow - allow other methods of finding index name"

PR #154, opened 2016-02-08
Faster time format parsing for most formats

ashb:
DateTime and Time have different APIs (DateTime is missing getutc) so making the time code support getting a Time or a DateTime would be rather complex (to deal with the UTC index vs not cases). However I was looking at the fluentd source code and in v0.14 they started using the [strptime gem](https://github.com/nurse/strptime): https://github.com/fluent/fluentd/blob/v0.14.0.pre.1/lib/fluent/parser.rb#L61-L71

That gem doesn't support all formats (for instance `%a` isn't supported) hence why they fallback to `Time.strptime` if creating the parser object fails. It's also fast:

```
(themisto tmp/fluent…search strptime:+)% ~/.rbenv/shims/bundle exec ruby bench.rb
                                     user     system      total        real
Time.parse                       7.910000   0.040000   7.950000 (  7.966731)
Time.strptime                    3.140000   0.020000   3.160000 (  3.176454)
Time.xmlschema                   2.880000   0.010000   2.890000 (  2.893675)
DateTime.strptime                1.460000   0.020000   1.480000 (  1.484857)
DateTime.strptime.to_time        2.810000   0.020000   2.830000 (  2.842479)
Strptime.strptime                0.130000   0.000000   0.130000 (  0.140551)
```

Benchmark code:

```ruby
require 'time'
require 'benchmark'
require 'strptime'
n = 200_000

Benchmark.bm(30) do |x|
  x.report('Time.parse') { n.times { Time.parse('Thu Nov 29 14:33:20 GMT 2001') } }
  x.report('Time.strptime') { n.times { Time.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y") } }
  x.report('Time.xmlschema') { n.times { Time.xmlschema("2001-11-29T09:33:20-05:00") } }
  x.report('DateTime.strptime') { n.times { DateTime.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y") } }
  x.report('DateTime.strptime.to_time') { n.times { DateTime.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y").to_time } }
  x.report('Strptime.strptime') { p = Strptime.new("%m/%d %H:%M:%S %z %Y"); n.times { p.exec("11/29 14:33:20 +00:00 2001") } }
end
```

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -3,6 +3,7 @@ require 'date'
 require 'excon'
 require 'elasticsearch'
 require 'uri'
+require 'strptime'
 
 class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   class ConnectionFailure < StandardError; end
@@ -17,6 +18,7 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :scheme, :string, :default => 'http'
   config_param :hosts, :string, :default => nil
   config_param :target_index_key, :string, :default => nil
+  config_param :time_key_format, :string, :default => nil
   config_param :logstash_format, :bool, :default => false
   config_param :logstash_prefix, :string, :default => "logstash"
   config_param :logstash_dateformat, :string, :default => "%Y.%m.%d"
@@ -42,14 +44,34 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
 
   def initialize
     super
+    @time_parser = _make_time_praser(@time_key_format)
   end
 
   def configure(conf)
     super
+    @time_parser = _make_time_praser(@time_key_format)
   end
 
   def start
     super
+  end
+
+  # once fluent v0.14 is released we might be able to use
+  # Fluent::Parser::TimeParser, but it doesn't quite do what we want - if gives
+  # [sec,nsec] where as we want something we can call `strftime` on...
+  def _make_time_praser(time_key_format)
+    if time_key_format
+      begin
+        # Strptime doesn't support all formats, but for those it does it's
+        # blazingly fast.
+        strptime = Strptime.new(time_key_format)
+        Proc.new { |value| strptime.exec(value).to_datetime }
+      rescue
+        Proc.new { |value| DateTime.strptime(value, time_key_format) }
+      end
+    else
+      Proc.new { |value| DateTime.parse(value) }
+    end
   end
 
   def client
@@ -156,18 +178,16 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
         target_index = record.delete @target_index_key
       elsif @logstash_format
         if record.has_key?("@timestamp")
-          time = Time.parse record["@timestamp"]
+          dt = @time_parser.call(record["@timestamp"]) rescue Time.at(time).to_datetime
         elsif record.has_key?(@time_key)
-          time = Time.parse record[@time_key]
+          dt = @time_parser.call(record[@time_key]) rescue Time.at(time).to_datetime
           record['@timestamp'] = record[@time_key]
         else
-          record.merge!({"@timestamp" => Time.at(time).to_datetime.to_s})
+          dt = Time.at(time).to_datetime
+          record.merge!({"@timestamp" => dt.to_s})
         end
-        if @utc_index
-          target_index = "#{@logstash_prefix}-#{Time.at(time).getutc.strftime("#{@logstash_dateformat}")}"
-        else
-          target_index = "#{@logstash_prefix}-#{Time.at(time).strftime("#{@logstash_dateformat}")}"
-        end
+        dt = dt.new_offset(0) if @utc_index
+        target_index = "#{@logstash_prefix}-#{dt.strftime(@logstash_dateformat)}"
       else
         target_index = @index_name
       end
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -3,7 +3,10 @@ require 'date'
 require 'excon'
 require 'elasticsearch'
 require 'uri'
-require 'strptime'
+begin
+  require 'strptime'
+rescue LoadError
+end
 
 class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   class ConnectionFailure < StandardError; end
@@ -67,6 +70,9 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
         strptime = Strptime.new(time_key_format)
         Proc.new { |value| strptime.exec(value).to_datetime }
       rescue
+        # Can happen if Strptime doesn't recognize the format; or
+        # if strptime couldn't be required (because it's not installed -- it's
+        # ruby 2 only)
         Proc.new { |value| DateTime.strptime(value, time_key_format) }
       end
     else
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -47,12 +47,12 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
 
   def initialize
     super
-    @time_parser = _make_time_praser(@time_key_format)
+    @time_parser = TimeParser.new(@time_key_format, @router)
   end
 
   def configure(conf)
     super
-    @time_parser = _make_time_praser(@time_key_format)
+    @time_parser = TimeParser.new(@time_key_format, @router)
   end
 
   def start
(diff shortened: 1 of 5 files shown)


ashb:
@pitr This doesn't use `DateTime.strptime` because dealing with `@utc_index` would make the code a bit gnarly (`DateTime` doesn't have ` `getutc` method or similar - you have to do `datetime.new_offset(0)` -- I could try making this work if you think it's worth it?

ashb:
Oh: 

> Gem::InstallError: strptime requires Ruby version ~> 2.0.


Fluent seems to support (or at least test) the following versions https://github.com/fluent/fluentd/blob/c8ad174b1d66f96220da528996732b8eb42bc56d/.travis.yml#L3-L9:

```yaml
rvm:
  - 2.0.0
  - 2.1
  - 2.2.3
  - 2.3.0
  - ruby-head
  - rbx
```

pitr:
Hmm, strptime gem looks super fast. Since fluentd 0.14 is not out yet, I don't think it's a good idea to remove support for 1.9 in this gem yet, but we can remove the dependency on strptime from gemspec, but still try to `require` it (falling back to existing implementation). This way if somebody wants a performance bump, they can 1) use ruby 2.0+, 2) install 'strptime' and 3) specify `time_format`.

Alternatively, we can try tackling the `DateTime.strptime` path. This way folks would just need to specify `time_format`. I know `DateTime.strptime.to_time` is basically the same speed as `Time.strptime` but it doesn't have to be if we never call `to_time`.

Thoughts? I can help with this if this is not as interesting to you @ashb since #153 is merged (thanks for that PR!)

ashb:
We should see what the performance of calling `datetime.new_offset(0)` is -- it might wipe out the benefit we get from using `Datetime.strptime`?

I like the idea of trying to require that gem and only using it if it's there. This is possibly the limit of what I'll be able to get done this week, so I'll re-work this PR to use that gem if it's installed, then falling back to `Time.parse`

I might not be able to get around to tyring a `DateTime.strptime` version this week as I've already spent about a week poking about in our fluentd config and tweaking it when I was meant to be doing other things ;)

ashb:
@pitr Here's a version that will use strptime gem if it's installed.

pitr:
Looks like `new_offset` is actually super fast:

```
>> td = DateTime.now
>> puts Benchmark.measure { 200_000.times { td.new_offset(0) } }
  0.060000   0.000000   0.060000 (  0.061781)
```

However,

```ruby
require 'time'
require 'benchmark'
require 'strptime'

n = 200_000

FORMAT = "%Y.%m.%d"

Benchmark.bm(20) do |x|
  x.report('Time.parse') { n.times { Time.parse('Thu Nov 29 14:33:20 GMT 2001').getutc.strftime(FORMAT) } }
  x.report('Time.strptime') { n.times { Time.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y").getutc.strftime(FORMAT) } }
  x.report('Time.xmlschema') { n.times { Time.xmlschema("2001-11-29T09:33:20-05:00").getutc.strftime(FORMAT) } }
  x.report('DateTime') { n.times { DateTime.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y").new_offset(0).strftime(FORMAT) } }
  x.report('Strptime') { p = Strptime.new("%m/%d %H:%M:%S %z %Y"); n.times { p.exec("11/29 14:33:20 +00:00 2001").getutc.strftime(FORMAT) } }
end
```

```
                           user     system      total        real
Time.parse             5.880000   0.070000   5.950000 (  5.959256)
Time.strptime          2.420000   0.040000   2.460000 (  2.458869)
Time.xmlschema         1.700000   0.030000   1.730000 (  1.733077)
DateTime               1.240000   0.040000   1.280000 (  1.278167)
Strptime               0.410000   0.010000   0.420000 (  0.419985)
```

Hmm, keeping it in DateTime is about 3 times slower than `strptime` gem. On the other hand, it's 5 times faster than current solution...

I personally think going with DateTime solution is the simplest, especially considering that this impacts only those who have `@timestamp` or `time_key` in their records.

@dterror thoughts?

ashb:
Now with passing tests, and using DateTime.strptime/DateTime everywhere.

Still currently uses strptime if it's installed. Latest benchmarks show that using Strptime and converting it to a DateTime object (to make the code uniform) is still 2x faster than using `DateTime.strptime`:

```
$ ~/.rbenv/shims/bundle exec ruby ./bench.rb
                           user     system      total        real
Time.parse             8.160000   0.090000   8.250000 (  8.313071)
Time.strptime          3.680000   0.040000   3.720000 (  3.819555)
Time.xmlschema         3.380000   0.020000   3.400000 (  3.420666)
DateTime               1.570000   0.010000   1.580000 (  1.590534)
Time.at.to_datetime    0.570000   0.000000   0.570000 (  0.576965)
Strptime               0.520000   0.010000   0.530000 (  0.530981)
Strptime.to_datetime   0.730000   0.000000   0.730000 (  0.729871)
```

Benchmark code:

```ruby
require 'date'
require 'benchmark'
require 'strptime'

n = 200_000

FORMAT = "%Y.%m.%d"
TIME = 1455019217

Benchmark.bm(20) do |x|
  x.report('Time.parse') { n.times { Time.parse('Thu Nov 29 14:33:20 GMT 2001').getutc.strftime(FORMAT) } }
  x.report('Time.strptime') { n.times { Time.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y").getutc.strftime(FORMAT) } }
  x.report('Time.xmlschema') { n.times { Time.xmlschema("2001-11-29T09:33:20-05:00").getutc.strftime(FORMAT) } }
  x.report('DateTime') { n.times { DateTime.strptime("Thu Nov 29 14:33:20 GMT 2001", "%a %b %d %H:%M:%S %Z %Y").new_offset(0).strftime(FORMAT) } }
  x.report('Time.at.to_datetime') { n.times { Time.at(TIME).to_datetime.new_offset(0).strftime(FORMAT) } }
  x.report('Strptime') { p = Strptime.new("%m/%d %H:%M:%S %z %Y"); n.times { p.exec("11/29 14:33:20 +00:00 2001").getutc.strftime(FORMAT) } }
  x.report('Strptime.to_datetime') { p = Strptime.new("%m/%d %H:%M:%S %z %Y"); n.times { p.exec("11/29 14:33:20 +00:00 2001").to_datetime.new_offset(0).strftime(FORMAT) } }
end
```

pitr:
Woa! That's an even better approach, and doing away with Time altogether. Looks really good. My only remaining concern is the name of the config. `time_format` feels too generic, I think it needs to be tied more with `time_key`. Maybe `time_key_format`? Otherwise, +1, :shipit:

ashb:
I went with `time_format` to mirror the [fluentd stock parsers](http://docs.fluentd.org/articles/parser-plugin-overview):

```
format /^(?<remote>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] "(?<method>\S+)(?: +(?<path>[^\"]*) +\S*)?" (?<code>[^ ]*) (?<size>[^ ]*)(?: "(?<referer>[^\"]*)" "(?<agent>[^\"]*)")?$/
time_format %d/%b/%Y:%H:%M:%S %z
```

Thought happy to change it if you think it's better as `time_key_format`.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -156,18 +184,16 @@ def write(chunk)
         target_index = record.delete @target_index_key
       elsif @logstash_format
         if record.has_key?("@timestamp")
-          time = Time.parse record["@timestamp"]
+          dt = @time_parser.call(record["@timestamp"]) rescue Time.at(time).to_datetime
```
dterror: I think silently rescuing to ingestion time is not super cool, let's log the error and let people know things are not as expected.

ashb: The reason I went this way is that it was closer to the previous behaviour - `Time.parse` is really quite permissive and accepts many many formats, and I'd personally rather have the logs but with the wrong timestamp than drop them  and log an error on the aggregator which might not get seen.

dterror: I meant fallback to ingestion time but log that you're doing it, don't drop the log.

ashb: Ahhh! right. Yeah I'll add that, good call.

ashb:
@pitr @dterror Just pushed – hopefully addressing both your comments. Let me know if you think the way I'm testing the error case is okay or if it might be too fragile?

dterror:
:+1: thanks @ashb

Issue #160, opened 2016-03-25
Make adding of "@timestamp" optional if time_key is specified


static-max:
I want to use Logstash-style index names, but my data to ingest already contains a timestamp field that I use for queries, and I don't want an additional field named "@timestamp" that occupies disk space and is never used.

Great plugin by the way!

pitr:
Sounds like a good idea, if you are not using Kibana. Feel free to contribute a PR.

static-max:
Ok, how should we name the config? `time_key_exclude_timestamp`?

pitr:
that sounds good

PR #161, opened 2016-04-06
Exclude timestamp

static-max:
Added a new config option `time_key_exclude_timestamp` to disable the copying of the value from field `time_key` to `@timestamp`. Works in both dynamic and non-dynamic.
See #160.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible


pitr:
awesomeness!

pitr:

PR #163, opened 2016-04-10
Change index name to lower-case

meetyg:
Change target_index to lower-case since Elasticsearch doesn't allow upper-case characters in index names.
Added some tests, and they are passing.
The plugin itself was checked by manually running it in fluentd, and it seems to work fine, changing upper-case index names to lowercase.

(check all that apply)
- [x] tests added
- [x] tests passing (not checked)
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -211,7 +211,11 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
       else
         target_index = @index_name
       end
-
+      
+      # Change target_index to lower-case since Elasticsearch doesn't
+      # allow upper-case characters in index names.
+      target_index = target_index.downcase
+      
       if @include_tag_key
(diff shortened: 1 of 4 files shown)


meetyg:
Done, please merge.

meetyg:
Sorry, still new to Github :-), didn't mean to close it.

pitr:
thanks, this is great!

Could you add that line to out_elasticsearch_dynamic.rb as well, probably somewhere [around here](https://github.com/meetyg/fluent-plugin-elasticsearch/blob/patch-1/lib/fluent/plugin/out_elasticsearch_dynamic.rb#L148) and I'll merge it

meetyg:
Added change to dynamic plugin too, as well as tests.
The build is OK, but I don't understand why the coverage went down.

pitr:
looks good, coveralls has always been weird

pitr:

meetyg:
That's great, thanks for being so cooperative :+1:

PR #168, opened 2016-04-19
Add support for flattening hashes

skwp:
Adds support for flattening elasticsearch hashes so that elasticsearch doesn't yell at you and drop events if you submit both an object and a concrete value to the same field, which is common in large applications gathering logs from many disparate events.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible


skwp:
I'm testing this right now. The unit tests are passing but the plugin doesn't seem to be flattening

pitr:
lgtm, let us know if you find why it's not flattening for you

skwp:
This is good to go. We're running live in production. We were screwing up in our tests unrelated to this plugin.

pitr:

PR #175, opened 2016-07-03
Added support for dot separated target_index_key and target_type_key

DocX:
Some docker plugins like Kubernetes or Docker metadata plugins adds nested objects that contains values usable for `target_type`. For example setting tag_key to `kubernetes.container-name` will use value in `record['kubernetes']['container-name']`. That is useful for `target_type` value because different containers may have different log formats (that can conflict with names in ES if send with same type)

This patch allows to specify dot separated path for target_index_key and target_type_key to make this metadata plugins usable for this scenario.

It is backwards compatible with non nested keys. It only adds small additional overhead if used.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ By default, the records inserted into index `logstash-YYMMDD` with UTC (Coordina
 
 ### target_index_key
 
-Tell this plugin to find the index name to write to in the record under this key in preference to other mechanisms.
+Tell this plugin to find the index name to write to in the record under this key in preference to other mechanisms. Key can be specified as path to nested record using dot ('.') as a separator.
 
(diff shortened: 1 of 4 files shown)


DocX:
Looks like new versions of some dependencies are breaking tests :-(

1. New version of `fluentd` itself was causing errors. Locked version `~>0.12.0` to make tests passing (the same version as was at the time of last master commit). But I know it is not amazing solution.

2. New version of this gem causing failure on 1.9.3: `Gem::InstallError: mime-types-data requires Ruby version >= 2.0.`

pitr:
Thanks for the PR. Please resolve conflict by merging in master, to fix issues not related to this PR (specifically take the change to gemspec from master)

DocX:
@pitr thanks, updated the patch from master.

pitr:
thanks @DocX

PR #189, opened 2016-08-18
Adds `remove_keys_on_update` feature

errm:
remove_keys_on_update

It is useful especially when using upsert to skip some keys
from the update document. This feature enables that.

We are using fluentd to load and update some datasets in es.
Another application then can update certain fields to mark the
status of a particular record. So we wan't the data to be loaded
with some initial values for these fields, and then on future
updates, not change these fields.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ spec/reports
 test/tmp
 test/version_tmp
 tmp
+.DS_Store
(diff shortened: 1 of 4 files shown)


pitr:
well, why not

pitr:
released with v1.6.0

Issue #33, opened 2014-03-25
Specify ElasticSearch index template


chschs:
I'm using fluentd with the in_syslog plugin and elasticsearch plugin to get syslog into elasticsearch, with a kibana frontend.

One of the problems I'm having though, is that the fields are indexed in elasticsearch so when I add a terms dashboard in kibana to give me, say, the top-10 hostnames, hostnames with dashes in them are broken up. so mysql-test-01 would come across as three hostnames: mysql, test, and 01.

Logstash got around this issue by making a "raw" version of several fields that is set to not-analyzed upon creation, so that you can run your dashboards against that instead.

More information here: http://www.elasticsearch.org/blog/logstash-1-3-1-released/

With syslog messages going into ES with this plugin, I'm finding that I'd like to have a "raw" or non-analyzed host (hostname) field and ident field (gives me the application). Unfortunately right now both of those fields are analyzed and it's messing with our dashboards.

stanhu:
+1 for making this built into the plugin.

This is the template used by logstash:

https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/elasticsearch-template.json

aleks-v-k:
There is a valuable reason to make index template support built into the plugin: in a containerized environment we do not know when elasticsearch container will start and we have to PUT index template to it before fluentd will send data. There are some possible workarounds but every of them looks really ugly. So **+1** for making built in support of index templates. @pitr

ssergiienko:
+1 for making built in support of index templates.

pitr:
To implement this behaviour, this gem would need to do the following additional work before writing a record to elasticsearch:

1. Check if the index exists
2. If not, create the index and the mapping

Seems like that might impact the performance. What do you think @aleks-v-k @ssergiienko? What about deleting old indices?

Can you provide more information about how you run ElasticSearch in "containerized environment"? Is this something Cloudlinux is working on (I'm not familiar with their offerings)?

stanhu:
The gem would only need to write the index templates once since a [wildcard match](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html) could be used. As seen in the [Logstash template](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/elasticsearch-template.json#L2), just adding something like `fluentd-*` should take care of things.

pitr:
[...]

PR #194, opened 2016-08-26
allow specification of a index template

aerickson:
Allow users to specify a template to define on connection creation (similar to logstash).

See https://github.com/uken/fluent-plugin-elasticsearch/issues/33.

- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -2,11 +2,14 @@
 require 'date'
 require 'excon'
 require 'elasticsearch'
+require 'json'
 require 'uri'
 begin
   require 'strptime'
 rescue LoadError
 end
+
+require_relative 'out_elasticsearch_template'
 
 class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   class ConnectionFailure < StandardError; end
@@ -49,8 +52,11 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :remove_keys_on_update_key, :string, :default => nil
   config_param :flatten_hashes, :bool, :default => false
   config_param :flatten_hashes_separator, :string, :default => "_"
+  config_param :template_name, :string, :default => nil
+  config_param :template_file, :string, :default => nil
 
   include Fluent::SetTagKeyMixin
+  include Fluent::ElasticsearchOutputTemplate
   config_set_default :include_tag_key, false
 
   def initialize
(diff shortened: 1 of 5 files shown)


aerickson:
@pitr Do you have any comments? I'd love to get this merged. Thanks!

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -143,6 +150,35 @@ def client
...
+      log.info("Template configured and already installed.")
+    end
+  end
+
+  def get_template
+    if !File.exists?(@template_file)
+      raise "If you specify a template_name you must specify a valid template file (checked '#{@template_file}')!"
+    end
+    file_contents = IO.read(@template_file).gsub(/\n/,'')
+    template = JSON.parse(file_contents)
```
pitr: simply leaving last line as `JSON.parse(file_contents)` is sufficient

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -77,6 +80,10 @@ def configure(conf)
     if @remove_keys_on_update && @remove_keys_on_update.is_a?(String)
       @remove_keys_on_update = @remove_keys_on_update.split ','
     end
+
+    if @template_name and @template_file
```
pitr: `&&` should be used instead of `and`

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -143,6 +150,35 @@ def client
...
+  def template_install(name, template_file)
+    if !template_exists?(name)
+      template_put(name, get_template())
+      log.info("Template configured, but no template installed. Installed '#{name}' from #{template_file}.")
+    else
+      log.info("Template configured and already installed.")
+    end
+  end
+
+  def get_template
```
pitr: `template_file` should be an argument

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -143,6 +150,35 @@ def client
...
+    end
+    file_contents = IO.read(@template_file).gsub(/\n/,'')
+    template = JSON.parse(file_contents)
+    return template
+  end
+
+  def template_exists?(name)
+    client.indices.get_template(:name => name)
+    return true
+  rescue
```
pitr: you need to specify which exceptions you are rescuing

pitr:
Looks good, I left a few comments. Could you also extract all template logic to a separate module.

aerickson:
@pitr Thanks for the feedback. Changes made.

I was rescuing the more specific exception earlier when I was testing, but I encountered an error about an undeclared exception class when fluent-plugin-aws-elasticsearch-service consumed fluent-plugin-elasticsearch... but that seems to be gone now (just tested). Not sure what changed. :)

pitr:
1.7.0 released with this change. Also, I renamed template code in e87598a since out_*.rb files are for plugins

PR #196, opened 2016-09-13
add param to specify multiple templates

se-m:
Multiple templates.

With `target_index`, `target_type` and `record_transformer` you can use only one output point for any number of indices.
It should be possible set multiple templates to work in such configs.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -54,6 +54,7 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
   config_param :flatten_hashes_separator, :string, :default => "_"
   config_param :template_name, :string, :default => nil
   config_param :template_file, :string, :default => nil
+  config_param :templates, :hash, :default => nil
 
   include Fluent::SetTagKeyMixin
   include Fluent::ElasticsearchIndexTemplate
@@ -87,6 +88,11 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
     if @template_name && @template_file
       template_install(@template_name, @template_file)
     end
+
+    if @templates
+      templates_hash_install (@templates)
+    end
+
   end
 
   def start
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -87,9 +87,7 @@ class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
 
     if @template_name && @template_file
       template_install(@template_name, @template_file)
-    end
-
-    if @templates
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -87,6 +88,11 @@ def configure(conf)
     if @template_name && @template_file
       template_install(@template_name, @template_file)
     end
+
+    if @templates
```
pitr: Either `template_name`/`template_file` should be used OR `templates`

test/plugin/test_out_elasticsearch.rb
```diff
@@ -138,6 +139,42 @@ def test_template_create_invalid_filename
...
+      to_return(:status => 404, :body => "", :headers => {})
+    #creation
+    stub_request(:put, "https://john:doe@logs.google.com:777/es//_template/logstash").
+      to_return(:status => 200, :body => "", :headers => {})
+    stub_request(:put, "https://john:doe@logs.google.com:777/es//_template/logstash1").
+      to_return(:status => 200, :body => "", :headers => {})
+    stub_request(:put, "https://john:doe@logs.google.com:777/es//_template/logstash2").
+      to_return(:status => 200, :body => "", :headers => {})
+
+    driver('test', config)
```
pitr: This test doesn't seem to have any assertions

lib/fluent/plugin/out_elasticsearch.rb
se-m: If you do so (either ... or) then the question arises:
What to give priority when there are both options? (Or just throw an error?)

test/plugin/test_out_elasticsearch.rb
se-m: I made by analogy with `test_template_create`. I think, that I should check at least,  that there was the right amount of requests , but I do not know how to do it. Can you advise me?

se-m:
@pitr 
I have an application with microservices architecture. The name of each tag built on the hierarchy:
MyApp.MyClientService.php-fpm.message.alert
MyApp.MyRestService.nodejs.message.info

Indices are dynamically defined by record_transformer using target_index.
Therefore it is convenient to have only one rule for output (e.g. MyApp.**). Due to the fact that the there is one only ouput rule, then i need to be able to define all the templates in one place.

pitr:
Sorry, took me some time to get to this PR. Please address comment and I'll merge

se-m:
@pitr 
fix tests. `templates` parameter now will be ignored, when `template_name` & `template_file` are set

pitr:
v1.8.0 is released with this change

PR #203, opened 2016-10-19
Optimize output plugins

repeatedly:
I sometimes receives the report which Elasticsearch plugin consumes lots of memory rathar than other plugins.
So I optimizes performance and memory usage by reducing object allocation.
Approaches are:
- Convert events into json before passes elasticsearch-ruby. elasticsearch-ruby re-creates incoming array in `bulk` method. Passing string avoids it.
- Reuse meta header object. Old implementation creates hash object in each loop. Each write operation uses corresponding fixed format, so we can reuse it.
- Use ObjectBufferedOutput plugin instead of BufferedOutput to reduce format/serialization cost.
- [x] tests passing
- [x] backward compatible

This patch changes lots of places, so need tests on some actual workloads.
On my laptop, memory usage becomes around 200MB from around 400MB with 1000,000 records import.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -9,6 +9,7 @@ begin
 rescue LoadError
 end
 
+require 'fluent/output'
 require_relative 'elasticsearch_index_template'
 
 class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -11,7 +11,7 @@ end
 
 require_relative 'elasticsearch_index_template'
 
-class Fluent::ElasticsearchOutput < Fluent::BufferedOutput
+class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
   class ConnectionFailure < StandardError; end
 
   Fluent::Plugin.register_output('elasticsearch', self)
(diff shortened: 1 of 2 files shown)


wyukawa:
Hi

I try to use this pull request.
I confirm that memory usage decreases significantly compared to 0.1.5.

Gemfile is the following
```
gem "fluent-plugin-elasticsearch",:git=>"https://github.com/repeatedly/fluent-plugin-elasticsearch.git", :ref => "b5b3a9c938fe97b46208b861c8ab72a3e24dc306"
```

* memory rss


* memory vss

wyukawa:
I use fluent-plugin-forest with fluent-plugin-elasticsearch.
I think that this is another reason to decrease memory after restarting.

pitr:
THANKS @repeatedly! I didn't even know about `ObjectBufferedOutput`.

I'd like somebody else to try this in their system and then I'll release this as alpha version

repeatedly:
Rebase for master changes.

repeatedly:
Downloading file and copy it into fluentd or td-agent gem is hard for almost users.
Google k8s team also tried this patch and no errors.
How about releasing dev version for easy to test?

pitr:
Released v1.9.0.rc.1 with this change

repeatedly:
Thanks!

wyukawa:
Thanks!

I upgrade to fluent-plugin-elasticsearch 1.9.0.rc.1 from 1.7.0 at after 10AM on November 8th。

The change of memory and cpu usage is the following.
Actually, memory and cpu usage decreases.

memory


cpu

repeatedly:
Thanks for long running test on your production.
The result was what I had expected :)

@pitr How about releasing v1.9.0?

pitr:
released 1.90, thanks again for this PR

PR #206, opened 2016-10-21
send method name is changed to send_bulk since it is a reserved keyword

avikam:
DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible


pitr:
Thanks, could you do the same to `out_elasticsearch_dynamic.rb`?

pitr:
v1.8.0 is released with this change

PR #220, opened 2016-12-02
Converting string to bool in dynamicconfig.

tanaka-takayoshi:
DESCRIPTION HERE
reload_connections and reload_on_failure are expected as bool in Elasticsearch Client.

There are many discussion to be related this issue.
https://discuss.elastic.co/t/elasitcsearch-ruby-raises-cannot-get-new-connection-from-pool-error/36252
https://forums.aws.amazon.com/thread.jspa?threadID=222600
https://github.com/atomita/fluent-plugin-aws-elasticsearch-service/issues/15

I found this PR will fix "Cannot get new connection from pool."  error when sending logs to some kind of elasticsearch clusters whose _nodes API doesn't return IP address of nodes. This kind of elasticsearch is not only AWS elasticsearch but also the elasticsearch behind http proxy.
Elasticsearch Ruby client expects reload_connections and reload_on_failure parameter as a bool type. 
https://github.com/elastic/elasticsearch-ruby/issues/164

However, when we use your dynamic configuration feature, these parameters would be string type. Then reload_connections method are invoked at some point even if we specify ``reload_connections false``. 
https://github.com/elastic/elasticsearch-ruby/blob/v2.0.0/elasticsearch-transport/lib/elasticsearch/transport/transport/base.rb#L70
```ruby
reload_connections!         if reload_connections && counter % reload_after == 0
```

If you find something wrong in my PR, please let me know.

(check all that apply)
- [ ] tests added: difficult to add test.
- [X] tests passing
- [X] README updated (if needed)
- [X] README Table of Contents updated (if needed)
- [X] History.md and `version` in gemspec are untouched
- [X] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch_dynamic.rb
+++ b/lib/fluent/plugin/out_elasticsearch_dynamic.rb
@@ -1,6 +1,12 @@
 # encoding: UTF-8
 require_relative 'out_elasticsearch'
 
+class String
+  def to_bool
+    self == 'true'
+  end
+end
+
 class Fluent::ElasticsearchOutputDynamic < Fluent::ElasticsearchOutput
 
   Fluent::Plugin.register_output('elasticsearch_dynamic', self)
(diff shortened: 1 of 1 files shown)


pitr:
Thanks! We are already doing `dynamic_config[..].to_i` in a few places, so this makes sense. Could you change your PR to do the following instead:

```ruby
class String
  def to_bool
    self == 'true'
  end
end

class Fluent::ElasticsearchOutputDynamic < Fluent::ElasticsearchOutput
# ...
```

This way you can call `dynamic_config[..].to_bool`, which is much cleaner

repeatedly:
Another way is using `Fluent::Config.bool_value`: https://github.com/fluent/fluentd/blob/master/lib/fluent/config/types.rb#L53

pitr:
What's the best way to use it? `include Fluent::Config` and `bool_value(@dynamic_config['reload_connections'])`? Can we use `Fluent::Configurable` for `dynamic_config` somehow?

repeatedly:
`Fluent::Config.bool_value(@dynamic_config['reload_connections'])` is enough.

pitr:
@tanaka-takayoshi could you make the change and I'll merge

tanaka-takayoshi:
@pitr I pushed the change.

pitr:
Thanks @tanaka-takayoshi @repeatedly

therc:
Any chance of a new release getting made with this change? I think we are affected. Having a tagged release would help. Thanks!

pitr:
Done, released in 1.9.1

Issue #222, opened 2016-12-07
Support nanosecond timestamps


yissachar:
Fluentd 0.14 has support for nanosecond timestamps but it seems that this plugin does not support that.

Example config:

```
<source>
  @type tail
  read_from_head true
  path catalina.out
  pos_file catalina.pos
  time_format %Y-%m-%d %H:%M:%S,%L
  tag foo.*
  format multiline
  format_firstline /\d{4}-\d{2}-\d{2}/
  format1 /^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}) \[(?<thread>.*)\] (?<level>[^\s]*)(?<message>.*)/
</source>

<match foo.**>
  @type elasticsearch
  logstash_format true
  flush_interval 5s
  host localhost
  port 9200
</match>
```

When using this config I end up with timestamps that are only resolved to the second. If I use stdout output instead of Elasticsearch then I see that the timestamps have millisecond granularity.

yissachar:
@pitr I started coding up the required changes and currently have something that works for me. However, supporting this feature requires moving to the v0.14 API, which means that users on Fluentd v0.12 would not be able to use new versions of this plugin until they upgrade their Fluentd version to >=0.14. Before I put any further work into cleaning up my code and submitting a PR, I'd like to know if this is something that you are OK with or not.

pitr:
@yissachar I am perfectly fine with that, 0.14 was released 6 months ago, about time we jump on that bandwagon. Also, I found [instructions on upgrading plugin from 0.12 to 0.14](http://docs.fluentd.org/v0.14/articles/plugin-update-from-v12) I am planning to follow section "Updating Plugins Overview" when releasing 0.14-only version of this plugin.

miketheman:
Bump. I filed myself a reminder to check back in on this issue since we notices our problem in January.

Is there anything we can do to help move this along? We'd really like the precision added via these timestamps.

ChipmunkV:
@yissachar, is there a WIP PR?

davidgacc:
I am trying to see the timestamp with miliseconds on Kibana. But it seems only to show only seconds

fluentd 0.14.19
td-agent 0.14.16

Example of my log:

2017-04-02 00:00:02,027 INFO  [com.***] (http--0.0.0.0) text

Ouput on Kibana:

April 2nd 2017, 00:00:02.000

`<source>
        @type tail
[...]

PR #223, opened 2016-12-08
Add support for nanosecond timestamps

yissacharcw:
Fixes https://github.com/uken/fluent-plugin-elasticsearch/issues/222

Unfortunately I couldn't run the tests due to some dependency issues in my local setup (sorry, I don't use Ruby, so don't have an environment setup).

I think the README section on [buffered output optons](https://github.com/uken/fluent-plugin-elasticsearch#buffered-output-options) should probably be tweaked, but I didn't know what to put there, so I left it as is.

(check all that apply)
- [x] tests added
- [x] tests passing
- [ ] backward compatible


yissacharcw:
The diff is a bit messy, would recommend looking at it with whitespace ignored (add `?w=1` to the GitHub diff URL).

fluent-plugin-elasticsearch.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = Gem::Requirement.new(">= 2.0".freeze)
 
-  s.add_runtime_dependency 'fluentd', '>= 0.10.43'
+  s.add_runtime_dependency 'fluentd', '>= 0.14.0'
```
cosmo0920: Fluentd 0.14.4, 0.14.5 have various bugs related to its built-in plugins.
Could you specify 0.14.8 or later here?
0.14.8 seems to be relatively stable.

yissacharcw: Updated, thanks.

yissacharcw:
@pitr I figured out how to get tests running locally and switched to the new API which made most of them pass, but there are still a few that are failing and I'm not sure why.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -12,357 +12,365 @@
...
-
-  include Fluent::ElasticsearchIndexTemplate
-
-  def initialize
-    super
-  end
+module Fluent::Plugin
+  class Fluent::ElasticsearchOutput < Output
+    class ConnectionFailure < StandardError; end
+
```
cosmo0920: Could you use `event_emitter` plugin here like this?

```ruby
helpers :event_emitter
```

This plugin helper creates `router` in current plugin context.

cosmo0920:
Keyword arguments of v0.14 test driver's `#run` may helps you:

https://github.com/fluent/fluentd/blob/ad481747ec6793432f38e3c34126d9dc9d63c385/lib/fluent/test/driver/base_owner.rb#L122

lib/fluent/plugin/out_elasticsearch.rb
yissachar: Thanks, though now I get a new error: `NoMethodError: undefined method `emit_error_handler' for nil:NilClass`

yissachar:
@cosmo0920 Can you expand on this? Which argument would help here?

lib/fluent/plugin/out_elasticsearch.rb
cosmo0920: This is because original `router` definition is not used in test.
If we want to test this case, it needs to adjust  for method definitions with `stub` or `mock` from `test-unit-rr`.

test/plugin/test_out_elasticsearch_dynamic.rb
```diff
@@ -10,8 +11,8 @@ def setup
     @driver = nil
   end
 
-  def driver(tag='test', conf='')
-    @driver ||= Fluent::Test::BufferedOutputTestDriver.new(Fluent::ElasticsearchOutputDynamic, tag).configure(conf)
+  def driver(conf='')
+    @driver ||= Fluent::Test::Driver::Output.new(Fluent::ElasticsearchOutputDynamic).configure(conf)
```
cosmo0920: It seems that `Fluent::ElasticsearchOutputDynamic` needs to be placed under `Fluent::Plugin` namespace.

cosmo0920:
Ah, I'm sorry. Above comment's keyword argument cannot help us.
Keywords arguments help us to stabilize test cases.
But this case does not fit for test driver's keyword argument.

yissachar:
@cosmo0920 @pitr I'm kind of out of my depth code. I don't know Ruby; this is the first Ruby code I've written. There are only a few failing tests but I don't have the necessary knowledge to make them pass.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -12,357 +12,365 @@
 require 'fluent/output'
```
pitr: This should be `require "fluent/plugin/output"` according to http://docs.fluentd.org/v0.14/articles/plugin-update-from-v12#buffered-output-plugins

yissachar:
@pitr @cosmo0920 I made the requested changes and rebased. Even fewer tests passing now... not sure why. Would greatly appreciate if either of you could take a close look at the failing tests and suggest what needs to be changed to make them pass.

equinox0815:
Hi,

First of all @yissacharcw thanks for your work. I justed successfully tested our setup with your patch applied and so far everything seems to be working. We are using td-agent 3.0.0 for ubuntu/xenial which contains fluentd 0.14.11. 
Besides using a non-stable version of fluentd do you think using the patch in the current state introduces a lot of new problems in a semi-production environment?

regards
 christian

yissachar:
@equinox0815 I don't see any issues with it. The changes I've made are exceedingly minimal. All I've done is update the fluentd version and made the necessary API changes to support v0.14. I plan on using this in production myself.

sampointer:
This overlaps somewhat with this issue: https://github.com/uken/fluent-plugin-elasticsearch/issues/235

jgoodall:
Is this going to be merged?

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -9,363 +9,373 @@
...
-  include Fluent::ElasticsearchIndexTemplate
-
-  def initialize
-    super
-  end
+module Fluent::Plugin
+  class ElasticsearchOutput < Output
+    class ConnectionFailure < StandardError; end
+
+    helpers :event_emitter
```
pitr: why is this needed?

yissachar: Hi @pitr, thanks for returning to this! I added this based on a review comment by @cosmo0920, previously we were getting errors `undefined local variable or method `router'`.

cosmo0920: In v0.14, `router` is not defined by default. It must write `helpers :event_emitter` to use `router` in v0.14 Output plugin.

dsapala:
Is there any current status on this?

yissachar:
Rebased against master. Still failing tests. Most failing tests are where I've used `driver.run` twice. It seems that's not the correct way to go about replacing the double `emits` that were previously there, but I'm not sure what the correct approach is. I've tried moving the `feed`s into a single `driver.run` but then tests fail with the incorrect number of tries.

Another puzzle is `log = driver.instance.router.emit_error_handler.log`. Obviously in v14 API that doesn't exist anymore, but I don't know what to repalce it with. I've tried `driver.instance.log`, but the test failed not finding the expected log there.

This PR would really benefit from a review from a Fluentd expert.

@repeatedly Is there any chance you could take a look at this?

repeatedly:
I don't have a time to review this patch.
@cosmo0920 Could you review it again?

cosmo0920:
Sure, I'll do it.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -9,365 +9,375 @@
...
+    config_param :template_name, :string, :default => nil
+    config_param :template_file, :string, :default => nil
+    config_param :templates, :hash, :default => nil
+    config_param :include_tag_key, :bool, :default => false
+    config_param :tag_key, :string, :default => 'tag'
+    config_param :time_parse_error_tag, :string, :default => 'Fluent::ElasticsearchOutput::TimeParser.error'
+    config_param :reconnect_on_error, :bool, :default => false
+
+    config_section :buffer do
+      config_set_default :chunk_keys, ['tag']
```
cosmo0920: How about specify default buffer type and use compat_parameters plugin helper?

```ruby
    helpers :event_emitter, :compat_parameters
    # ...
    DEFAULT_BUFFER_TYPE = "memory"
    # ...
    config_section :buffer do
      config_set_default :@type, DEFAULT_BUFFER_TYPE
      config_set_default :chunk_keys, ['tag']
    end
    
    def configure(conf)
      compat_parameters_convert(conf, :buffer)
      # ...
    end
```

test/plugin/test_out_elasticsearch.rb
```diff
@@ -789,8 +827,9 @@ def test_uses_no_subsecond_precision_by_default
     rescue
       time = Fluent::Engine.now
     end
-    driver.emit(sample_record, time)
-    driver.run
+    driver.run(default_tag: 'test') do
+      driver.feed(time.to_i, sample_record)
+    end
     assert(index_cmds[1].has_key? '@timestamp')
     assert_equal(index_cmds[1]['@timestamp'], Time.at(time).iso8601)
```
cosmo0920: Default value of `@time_precision` is 9.
I guess that 

```ruby
assert_equal(index_cmds[1]['@timestamp'], Time.at(time).iso8601(9))
```

is correct assertion.

test/plugin/test_out_elasticsearch.rb
```diff
@@ -754,8 +790,9 @@ def test_uses_custom_time_key_format_logs_an_error(tag_for_error)
     ts = "2001/02/03 13:14:01,673+02:00"
     index = "logstash-#{Date.today.strftime("%Y.%m.%d")}"
 
-    driver.emit(sample_record.merge!('@timestamp' => ts))
-    driver.run
+    driver.run(default_tag: 'test') do
+      driver.feed(sample_record.merge!('@timestamp' => ts))
+    end
 
     log = driver.instance.router.emit_error_handler.log
```
cosmo0920: As mentioned above,  `driver.instance.router.emit_error_handler.log` is gone away in v0.14.
How about use `flexmock` to mock `emit_error_event` like this?

```ruby
flexmock(driver.instance.router).should_receive(:emit_error_event)
```

test/plugin/test_out_elasticsearch.rb
```diff
@@ -954,10 +1001,10 @@ def test_connection_failed_retry
       raise Faraday::ConnectionFailed, "Test message"
     end
 
-    driver.emit(sample_record)
-
-    assert_raise(Fluent::ElasticsearchOutput::ConnectionFailure) {
-      driver.run
+    assert_raise(Fluent::Plugin::ElasticsearchOutput::ConnectionFailure) {
```
cosmo0920: In v0.14, driver#feed does not raise Error.
We can remove this assertion safely.

test/plugin/test_out_elasticsearch.rb
```diff
@@ -974,14 +1021,17 @@ def test_reconnect_on_error_enabled
     end
```
cosmo0920: test_reconnection_on_error_enabled test case is affected to test driver behavior change.

I tried to adopt this test case for v0.14 test driver as follows:

```ruby
  def test_reconnect_on_error_enabled
    connection_resets = 0

    stub_elastic_ping(url="http://localhost:9200").with do |req|
      connection_resets += 1
    end

    stub_request(:post, "http://localhost:9200/_bulk").with do |req|
      raise ZeroDivisionError, "any not host_unreachable_exceptions exception"
    end

    driver.configure("reconnect_on_error true\n")

    assert_raise(ZeroDivisionError) {
      driver.run(default_tag: 'test', flush: true, wait_flush_completion: false, force_flush_retry: false, shutdown: false) do
        driver.feed(sample_record)
      end
    }

    driver.run(default_tag: 'test', flush: true, wait_flush_completion: false, force_flush_retry: true, shutdown: false) do
      driver.feed(sample_record)
    end

    # XXX: connection_resets should be 2. How to fix it?
    assert_equal(connection_resets, 1)
  end
```

But I could not obtain same value of `connection_resets` in v0.12.
How do we treat this test case?

cosmo0920:
Any response?

lib/fluent/plugin/out_elasticsearch.rb
yissachar: Good idea!

test/plugin/test_out_elasticsearch.rb
yissachar: Good catch.

test/plugin/test_out_elasticsearch.rb
yissachar: Yes, this is the main problem that I can't figure out how to address.

test/plugin/test_out_elasticsearch.rb
yissachar: Hmm, I'm not sure what you mean. The code still throws an error when using it with a host that can't be reached. So shouldn't we have a test that verifies this?

test/plugin/test_out_elasticsearch.rb
yissachar: I don't really understand well enough what this test is actually testing, and how your proposed code interacts.

yissachar:
@cosmo0920 Really sorry for the delay! I've pushed some fixes, and left some comments. Thanks for your review!

test/plugin/test_out_elasticsearch.rb
yissachar: I see the exception is being thrown, just not where we are catching it: https://travis-ci.org/uken/fluent-plugin-elasticsearch/jobs/252927493#L215-L235

Not really sure why this is happening.

yissachar: So I think I understand now, the driver catches the exception and prints it, as seen here: https://github.com/fluent/fluentd/blob/master/lib/fluent/test/driver/base.rb#L141-L150. I don't understand why this behaviour is desirable, or how we can work around it.

yissachar: I guess the real question is why the driver is waiting until `after_shutdown` to flush, instead of flushing immediately.

yissachar: I was able to get the test to pass by flushing immediately, not shutting down, and sleeping for as long as it takes all the retries to go throw:

```
assert_raise(Fluent::Plugin::ElasticsearchOutput::ConnectionFailure) {
  driver.run(default_tag: 'test', flush: true, shutdown: false) do
    driver.feed(sample_record)
  end
  sleep 6
}
```

Does this seem right to you @cosmo0920 ?

test/plugin/test_out_elasticsearch.rb
cosmo0920: This test case should confirm whether or not occurring error events.
In v0.12, we can use `driver.instance.router.emit_error_handler.log` to confirm error events.
But, in v0.14, `driver.instance.router.emit_error_handler.log` is removed.
We should another way to confirm error occurred.

In the above my code snippets, I tried to confirm `emit_error_event` method invocation with mock.

cosmo0920: Umm, but I could not find how to fix this test properly in v0.14, though.... :<

test/plugin/test_out_elasticsearch.rb
cosmo0920: No, I don't.
Hard-coded magic number causes unstable test.
Instead, how about remove assert_raise in this case?
Because this test checks the counts of connection resets.

> So I think I understand now, the driver catches the exception and prints it, as seen here: https://github.com/fluent/fluentd/blob/master/lib/fluent/test/driver/base.rb#L141-L150. I don't understand why this behaviour is desirable, or how we can work around it.

Also I have no idea why fluentd v0.14 test driver choices current behavior.
I think that we should send feature request to https://github.com/fluent/fluentd if current behavior is undesirable.

> I guess the real question is why the driver is waiting until after_shutdown to flush, instead of flushing immediately.

Because v0.14 driver supports `try_write` (asynchronous write) feature.
If flushing immediately, it cannot support asynchronous write method testing.

test/plugin/test_out_elasticsearch.rb
cosmo0920: @pitr Could you ask why this test case using `driver.instance.router.emit_error_handler.log` to assert error event log?
I think that it is too complicated to convert v0.14 style test case.
  
In v0.14, `driver.instance.router.emit_error_handler.log` is completely gone away and no equivalent method.

cosmo0920:
@yissachar I've commented to your questions and changes.

@pitr some test cases is very complicated to fix such as using `driver.instance.router.emit_error_handler.log`. How do we fix these cases?

test/plugin/test_out_elasticsearch.rb
yissachar: Thanks, your comments explained what is going on. I've added a new test using `flexmock` to verify that the error is being emitted.

test/plugin/test_out_elasticsearch.rb
yissachar: I understand your point regarding hard-coded sleep. The problem is that the retry logic incorporates a backoff, and it seems even when specifying immediate flush that it somehow does not wait automatically for the sleep from the retry.

I don't think removing the assertion is correct, because then we will be missing half the logic of the test. We will know that it tried to reconnect 3 times, but we don't know that after that a ConnectionFailure is thrown. I guess this is not the end of the world, since it is sort of implied.

davidgacc:
@yissacharcw this patch of nanoseconds timestamp has already merged ?
This patch works with Fluentd 0.14.19  ?

yissachar:
@davidgacc No this is not merged, but you can use it anyway. Yes, I've tested it working with Fluentd 0.14.19

cosmo0920:
@gihad What is the status of this plugin maintenance?
It seems that almost ten of PRs and tens of issues are still opened since May 2017.
Could you solve issues and merge PRs?

If you have not enough time to do, I want to maintain this plugin.
So, please add me to collaborator in this repository?
(I'm one of fluent organization member and I know how to migrate to use v0.14 API)

gihad:
@cosmo0920 Will add you as a maintainer. Thank you.

gihad:
It's done, cosmo0920 is now a maintainer.

cosmo0920:
Thank you for your invitation.
I accepted it.

cosmo0920:
One more question.
Do you want to release into rubygems by yourself, @gihad?
If you want to completely take over this project in near future, please invite me to rubygems owner.

cosmo0920:
Hi all, I added commits to pass test cases for v0.14.
I have no idea how to treat a few test cases for now.
But we got reached all test passing in v0.14 migration work. :muscle:

Thanks for great work @yissachar! Your work is very helpful for us. :smile:

gihad:
@cosmo0920 I haven't pushed this to rubygems, yet the last 2 versions (1.9.4, 1.9.5) I "released" made their way there: https://rubygems.org/gems/fluent-plugin-elasticsearch/versions/

I checked CI (Travis) and I don't see it doing any pushes. I suspect either @pitr or @dterror (They worked at Uken a few years ago) pushed it manually.

@pitr or @dterror Can youadd both gihad and cosmo0920 to the project in rubygems? Thanks.

yissachar:
@cosmo0920 Awesome, thanks for stepping up to become maintainer!

cosmo0920:
I'll merge this PR when joining rubygems fluent-plugin-elasticsearch project.
And then, master branch will be aimed to work with Fluentd v0.14.8(td-agent3) or later.

I created v0.12 branch which is aimed to backport feature for 1.9.x which can work with Fluentd v0.12(td-agent2).

Tom-Fawcett:
@gihad @cosmo0920 FYI for 1.9.4 and 1.9.5 I emailed @pitr using his rubygems profile address.

gihad:
Thanks, that confirms my theory above. I will try contacting @pitr externally, if I can't get access we can just start a new rubygems gem for version 2.

gihad:
@cosmo0920 and I have access to rubygems gem now.

cosmo0920:
I've merged this PR.
I'm preparing 2.0.0.rc1 release.

cosmo0920:
I'd published as 2.0.0.rc.1.

nsaud01:
How do I actually get milliseconds or nanoseconds in timestamps to work? Is there any documentation on this?

cosmo0920:
To get nanoseconds on timestamp, you can use `%N` in time format: https://ruby-doc.org/stdlib-2.3.1/libdoc/date/rdoc/DateTime.html#strftime-method

And you can use [time_key_format](https://github.com/uken/fluent-plugin-elasticsearch/blob/2eab42ab989d8070c38508d47007fb08c7353a4f/lib/fluent/plugin/out_elasticsearch.rb#L34) parameter to get nanosecond precision. 

For example,

```aconf
time_key_format %Y-%m-%d %H:%M:%S.%N%z
```

will handle nanoseconds and timezone.

nsaud01:
Thanks @cosmo0920. I've tried that time_key_format in my fluentd.conf file to no avail. Where in the setup should it be? Here is my conf file:
```
# fluentd/conf/fluent.conf
<source>
  @type forward
  port 24224
  bind 0.0.0.0
</source>
<match *.**>
  @type copy
  <store>
    @type elasticsearch
    host elasticsearch
    port 9200
    logstash_format true
    logstash_prefix fluentd
    logstash_dateformat %Y%m%d
    include_tag_key true
    type_name access_log
    tag_key @log_name
    flush_interval 1s
  </store>
  <store>
    @type stdout
  </store>
</match>
```

cosmo0920:
@nsaud01 Could you confirm `@timestamp` field which includes milliseconds or nanoseconds precision time string in payload?
If `@timestamp` is included in payload, elasticsearch plugin will parse and handle millisecond, microsecond, or nanosecond precision time:

Sending log:

```log
'{"@timestamp":"2017-08-19 12:00:15.121T+09:00", "version":{"unstable":0.14, "stable":0.12}}'
```

will be handled in elasticsearch:

```json
      {
        "_index": "fluentd-20170819",
        "_type": "access_log",
        "_id": "AV6X45mI11oUNWVkG5uY",
        "_score": 1,
        "_source": {
          "@timestamp": "2017-08-19 12:00:15.121T+09:00",
          "version": {
            "unstable": 0.14,
            "stable": 0.12
          },
          "@log_name": "my.logs"
        }
      },
```

nsaud01:
@cosmo0920 The @timestamp is showing no milliseconds either. I'm using the fluentd docker logging driver to forward logs to my fluentd instance. Could the logging driver be v0.12 and not parsing the time? Here is my new config file:

````
<source>
  @type forward
  port 24224
</source>
<filter **>
  @type concat
  key log
  stream_identity_key container_id
  multiline_start_regexp /^\S+/
  multiline_end_regexp /\s+.*more$/
  flush_interval 120s
  timeout_label @processdata
</filter>
<label @ERROR>
  <match **>
    @type stdout
  </match>
</label>
<label @processdata>
  <match **>
    @type stdout
  </match>
</label>
<match **>
  @type elasticsearch
  host es-domain.us-east-1.es.amazonaws.com
  port 443
  scheme https
  ssl_verify false
  logstash_format true
  logstash_prefix test-logs
  flush_interval 10s
  request_timeout 120s
  type_name fluentd
  utc_index true
  time_key_format %Y-%m-%dT%H:%M:%S.%N%z
  include_tag_key true
  resurrect_after 5s
  reload_connections false
  buffer_chunk_limit 8m
</match>
````

And here is my docker daemon config:

````
 {
   "log-driver": "fluentd",
   "log-opts": {
     "fluentd-address": "fluent-host:24224"
   }
 }
````

wardviaene:
The issue https://github.com/moby/moby/issues/17181 still needs to be resolved before you'll get your messages from docker in sub-second format. The only solution now is to put a new timestamp on the incoming record and write it in subsecond format to ES

timestamp_ms ${ require 'time'; Time.now.utc.iso8601(3) }
time_key timestamp_ms

You possibly also need time_precision and time_key_format

nsaud01:
Thanks @wardviaene! Do you have an example of implementing that solution by chance?

Issue #193, opened 2016-08-24
The latest version breaks with usage of @type elasticsearch_dynamic


pragmaticivan:
#### Problem

I'm testing the latest release with fluentd 0.14, if I use @type elasticsearch_dynamic even If I don't use any dynamic value it still breaks.
#### Steps to replicate

```
<source>
  type forward
  port 24224
  bind 0.0.0.0
</source>

<filter docker.apache.*>
  @type record_transformer

  enable_ruby true
  renew_record true
  keep_keys source

  <record>
    containerUid ${record['container_id']}
    recorded_json ${record.to_json}
    level 3
  </record>
</filter>

<match docker.apache.*>
  @type elasticsearch_dynamic
  host elasticsearch
  port 9200

  logstash_format true
  logstash_prefix "fluentd-docker1t"
  time_key_exclude_timestamp true

  buffer_type memory
  buffer_chunk_limit 1m
  buffer_queue_limit 128
  flush_interval 1s
  retry_wait 1s
  max_retry_wait 10s
</match>
```

Error: 

```
fluentd          | 2016-08-24 18:50:02 +0000 [info]: listening fluent socket on 0.0.0.0:24224
fluentd          | 2016-08-24 18:50:09 +0000 [warn]: failed to flush the buffer. plugin_id="object:2aac5010251c" retry_time=0 next_retry=2016-08-24 18:50:10 +0000 chunk="53ad5c0187743d67a775301e39b5c61b" error_class=NoMethodError error="undefined method `+' for nil:NilClass"
fluentd          |   2016-08-24 18:50:09 +0000 [warn]: /home/ubuntu/ruby/lib/ruby/gems/2.2.0/gems/excon-0.52.0/lib/excon/connection.rb:228:in `request'
fluentd          |   2016-08-24 18:50:09 +0000 [warn]: /home/ubuntu/ruby/lib/ruby/gems/2.2.0/gems/faraday-0.9.2/lib/faraday/adapter/excon.rb:55:in `call'
fluentd          |   2016-08-24 18:50:09 +0000 [warn]: /home/ubuntu/ruby/lib/ruby/gems/2.2.0/gems/faraday-0.9.2/lib/faraday/rack_builder.rb:139:in `build_response'
[...]

PR #224, opened 2016-12-20
Fix elasticsearch_dynamic for v0.14

repeatedly:
`elasticsearch_dynamic` depends on fluentd's internal implementation and it is why `elasticsearch_dynamic` doesn't work with v0.14.
This patch fixes the problem by changing the approach.

- Use whitelist approach for dynamic parameters.
- Use fixed value for connection related parameters. These are not dynamic.
- Add v0.12 test for compatibility check in travis

I confirmed https://github.com/uken/fluent-plugin-elasticsearch/issues/193#issuecomment-242493899 this configuration works.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,9 @@ rvm:
   - 2.2
   - 2.3.1
 
+gemfile:
+ - Gemfile
+ - Gemfile.v0.12
+
 script: bundle exec rake test
 sudo: false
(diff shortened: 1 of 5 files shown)


therc:
This worked for me. Thanks, Masa!

therc:
Any chance this could get reviewed/merged? It's the last blocker for us to switch from 0.14.0.pre.1.

pitr:
Merged, released with 1.9.2. Thanks @repeatedly so much!!

Issue #227, opened 2016-12-30
mapper_parsing_exception not communicated through logs


zgolus:
When trying to index a document which would cause mapping type conflict (e.g. when trying to store a string to a field declared as an object) I get no indication of failure from the plugin, it just silently skips the document, however when trying to index the same document directly in ES I'm getting mapper_parsing_exception (400). I would expect in such situation that a log entry is created so the issue can be spotted and fixed.

I'm using dockerized fluentd v0.12.29 with fluent-plugin-elasticsearch v1.9.0 and fluent-plugin-aws-elasticsearch-service v0.1.6 on AWS Elasticsearch Service (2.3).

Best,
Jakub

pitr:
Please provide steps to replicate. You can use this gist as a starting point

zgolus:
I have created a gist with all files needed to recreate the issue: https://gist.github.com/zgolus/813ea2c30c142b9c740f50529a709878

To reproduce the issue please do the following:

Clone the gist:
`git clone https://gist.github.com/813ea2c30c142b9c740f50529a709878.git`
`cd 813ea2c30c142b9c740f50529a709878`

Build the image:
`docker build -t ps-fluentd .`

Start docker compose which will spawn Fluentd, Elasticsearch, Kibana and Nginx which will produce some logs:
`docker-compose up`

Deliberately preconfigure Elasticsearch with a mapping that is wrong:
`http PUT http://localhost:9200/fluentd < mapping.json`

Hit NGINX container to produce a log that doesnt match the mapping that we just provided:
`http GET http://localhost:80`

Make sure that no log was added to "fluentd" index:
`http GET http://localhost:9200/_cat/count/fluentd`

Check docker compose output and see that Elasticsearch reported an error regarding wrong mapping:
[...]

PR #230, opened 2017-01-03
add additional logging when client.bulk results in errors

zgolus:
Additional logging when client.bulk results in errors (https://github.com/uken/fluent-plugin-elasticsearch/issues/227).

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible

The tests are running after the changes and I have checked that the logging actually happens in my setup (described in the issue report). I have also added the change to the dynamic mode but I haven't tested it, however according to unit tests it should be fine.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -350,7 +350,12 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
   def send_bulk(data)
     retries = 0
     begin
-      client.bulk body: data
+      response = client.bulk body: data
+      if response['errors']
+        message = "Could not push log to Elasticsearch.\n"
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -350,7 +350,12 @@ def get_parent_of(record, path)
   def send_bulk(data)
     retries = 0
     begin
-      client.bulk body: data
+      response = client.bulk body: data
+      if response['errors']
+        message = "Could not push log to Elasticsearch.\n"
```
pitr: Multiline logs aren't fun to parse, let's simplify this to `"Could not push log to Elasticsearch: #{response}"` so as much information is available as can be for debug purposes

zgolus: Good point. I will fix it.

PR #239, opened 2017-02-07
Allow overwriting existing index template

chancez:
Basically the same as https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-template_overwrite

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -67,6 +67,7 @@ module Fluent::Plugin
     config_param :flatten_hashes_separator, :string, :default => "_"
     config_param :template_name, :string, :default => nil
     config_param :template_file, :string, :default => nil
+    config_param :template_overwrite, :bool, :default => false
     config_param :templates, :hash, :default => nil
     config_param :include_tag_key, :bool, :default => false
     config_param :tag_key, :string, :default => 'tag'
@@ -114,9 +115,9 @@ module Fluent::Plugin
       end
 
       if @template_name && @template_file
-        template_install(@template_name, @template_file)
+        template_install(@template_name, @template_file, @template_overwrite)
       elsif @templates
-        templates_hash_install (@templates)
+        templates_hash_install(@templates, @template_overwrite)
       end
 
       @meta_config_map = create_meta_config_map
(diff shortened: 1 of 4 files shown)


chancez:
I'd be glad to write a test for this, I just need a bit of guidance, it doesn't seem like existing tests verify that the old behavior didn't update the template if it already exists, which I'd need to add, and then another test to validate it is updated when the overwrite setting is enabled.

chancez:
Huzzah. Got test coverage back up for the new functionality, and improved the existing tests to assert the existing behavior as well. Let me know if you think I'm missing some tests.

chancez:
@repeatedly Are you able to review this or does @pitr the only person who can merge/review things for this project?

repeatedly:
@chancez I can't review because I'm not familiar with ES itself.
AFAIK, @pitr and @dterror are main maintainers but I'm not sure how they have a time to review all patches.
If they can, adding one more maintainer who knows ruby and fluentd is one approach.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -55,6 +55,7 @@ class ConnectionFailure < StandardError; end
   config_param :flatten_hashes_separator, :string, :default => "_"
   config_param :template_name, :string, :default => nil
   config_param :template_file, :string, :default => nil
+  config_param :template_overwrite, :string, :default => false
```
pitr: should be `:bool`

chancez: Ah, good catch. Fixed.

chancez:
@pitr Yep. That is the expected behavior, and also what logstash does. Generally there will be no conflicts and overwriting at boot up is fine since most the time the template will be the same across all instances of fluentd.

chancez:
@pitr Can we merge this? Anything outstanding?

dannyk81:
@pitr @cosmo0920 just wondering where this stands?

chancez:
I just rebased. It should be good to go.

cosmo0920:
I'd appreciate if you sent back port PR into v0.12 branch.
Thanks!

chancez:
@cosmo0920 as a separate PR?

cosmo0920:
@chancez Yes. Could you send a separate PR into v0.12 branch?

cosmo0920:
I'm planning to merge and release this PR as v2.3.0.
And then, I'm also planning to backport and release as v1.13.0.

cosmo0920:
I've release this as v2.3.0.
Thank you for your contribution, @chancez !

chancez:
Thanks @cosmo0920.

PR #249, opened 2017-03-28
configurable sub-second precision with no time key

sampointer:
For input plugins that do not provide a time key as part of the
record but do provide a `time` to the router, allow the degree
of sub-second time precision to be configurable.

Some sources (such as AWS CloudWatch) may provide an accurate time
source but do not include a time portion for an otherwise free-form
record: there is nothing to parse.

In this case the casting of a `DateTime` to a `String` loses any
sub-second precision.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Note: For Amazon Elasticsearch Service please consider using [fluent-plugin-aws-
   + [logstash_prefix](#logstash_prefix)
   + [logstash_dateformat](#logstash_dateformat)
   + [time_key_format](#time_key_format)
+  + [time_precision](#time_precision)
   + [time_key](#time_key)
   + [time_key_exclude_timestamp](#time_key_exclude_timestamp)
   + [utc_index](#utc_index)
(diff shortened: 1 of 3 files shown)


sampointer:
Hi! Is there anything I can do to help expedite this PR?

Tom-Fawcett:
@sampointer how do you think this compares to #223?

sampointer:
They're solving two different problems.

https://github.com/uken/fluent-plugin-elasticsearch/pull/223/ is basically doing "fluentd API change stuff" for plugins, which basically amounts to moving some stuff around in the class hierarchy. fluentd have stuffed semver for their plugin API: https://github.com/fluent/fluentd/issues/1449

This fixes a specific bug (and it is a bug) with a type cast losing time granularity in a specific situation.

They should both be merged.

Tom-Fawcett:
@sampointer thanks for your explanation, I was just curious due to a specific similarity between your pull and #223 -

#223
```
+          record[TIMESTAMP_FIELD] = dt.iso8601(9).to_s
```
#249
```
+          record[TIMESTAMP_FIELD] = dt.iso8601(@time_precision)
```

Anyway, regarding getting this merged; @gihad 's comment on #250 explains that there is unfortunately no active maintainer at this time.

gihad:
@sampointer I took a look at this PR - seems backwards compatible and everything looks good to me. Thanks for the contribution.

Tom-Fawcett:
@gihad thanks for merging this one.

I notice that https://rubygems.org/gems/fluent-plugin-elasticsearch is still only showing 1.9.3 and not the two newer versions. Is that something you can resolve?

PR #264, opened 2017-06-30
add the parsed time to records

harryge00:
DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

Fix the bug that the parsed time is not added into the record.

The original code just get the time from `record` and add it back as `@timestamp`. The parsed time `dt` has not been used!
```
           rts = record[@time_key]
           dt = parse_time(rts, time, tag)
           record[TIMESTAMP_FIELD] = rts unless @time_key_exclude_timestamp
```
Say my record is `{"a":"123", "time":"2017-06-30T06:25:05.197419064Z"}`. After processing, it just becomes `{"a":"123", "time":"2017-06-30T06:25:05.197419064Z", "@timestamp": "2017-06-30T06:25:05.197419064Z"}`.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -298,7 +298,7 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
         elsif record.has_key?(@time_key)
           rts = record[@time_key]
           dt = parse_time(rts, time, tag)
-          record[TIMESTAMP_FIELD] = rts unless @time_key_exclude_timestamp
+          record[TIMESTAMP_FIELD] = dt.strftime(@time_key_format ? @time_key_format : String.new) unless @time_key_exclude_timestamp
         else
           dt = Time.at(time).to_datetime
           record[TIMESTAMP_FIELD] = dt.iso8601(@time_precision)
@@ -308,7 +308,6 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
       else
         target_index = @index_name
       end
-
       # Change target_index to lower-case since Elasticsearch doesn't
       # allow upper-case characters in index names.
       target_index = target_index.downcase


meetyg:
Hello. 
I had the same problem that your PR should fix, but I see that it hasn't been merged.
Maybe because the build failed on a unit test named `test_uses_custom_time_key`.
The test fails because by default `@time_key_format` parameter is nil, and you use it in ` record[TIMESTAMP_FIELD] = dt.strftime(@time_key_format) ...`

After looking for the reason of the fail, I'm suggesting that the fix you did should check whether  `@time_key_format` parameter is supplied or not. 
Line 301 should look like this:
`record[TIMESTAMP_FIELD] = dt.strftime(@time_key_format ? @time_key_format : String.new) unless @time_key_exclude_timestamp`.

Since you allready dealt with this issue, and probably need this fix too, I would appreciate it if you could redo the fix as I suggested and get the PR merged.
Thanks

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -298,7 +298,7 @@ def write_objects(tag, chunk)
         elsif record.has_key?(@time_key)
           rts = record[@time_key]
           dt = parse_time(rts, time, tag)
-          record[TIMESTAMP_FIELD] = rts unless @time_key_exclude_timestamp
+          record[TIMESTAMP_FIELD] = dt.strftime(@time_key_format ? @time_key_format : String.new) unless @time_key_exclude_timestamp
```
cosmo0920: Could you add test case to verify this?

meetyg:
Sorry, my bad...
I thought that strftime would use its default format if it received an a empty string, but it actually treats is as format.
The existing unit tests are good and should cover this change (including `test_uses_custom_time_key_format` which checks the same case but with the format provided).

I am editing my suggestion for the fix.

meetyg:
I opened PR #285 for a new unit test that should cover the case when both `time_key` and `time_key_format` are provided.

PR #271, opened 2017-07-18
WIP: Allow logstash prefix via key in record

portante:
DESCRIPTION HERE

This is a Work-in-Progress to test out a simple behavior change to obviate the need for elasticsearch_dynamic in OpenShift's origin-aggregated-loggings fluentd use.

Basically, we only need to be able to specify the index prefix dynamically. This can be provided by a simple key in the record being indexed.

It avoids the overhead CPU calculations necessary to eval all the dynamic parameters.

This is untested, and purely a proof of concept to demonstrate a possibility.

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -14,6 +14,11 @@ require_relative 'elasticsearch_index_template'
 
 class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
   class ConnectionFailure < StandardError; end
+  class BulkIndexQueueFull < StandardError; end
+  class ElasticsearchOutOfMemory < StandardError; end
+  class ElasticsearchVersionMismatch < StandardError; end
+  class UnrecognizedElasticsearchError < StandardError; end
+  class ElasticsearchError < StandardError; end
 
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -88,6 +89,10 @@ def configure(conf)
       @remove_keys_on_update = @remove_keys_on_update.split ','
     end
 
+    if @logstash_prefix_key && @logstash_prefix_key.is_a?(String)
+      @logstash_prefix_key = @logstash_prefix_key.split '.'
```
richm: What does this do?

portante: Not sure exactly, just copied what `target_index_key` did.

richm: `@target_index_key` can be in the form "parentkey.childkey" https://github.com/uken/fluent-plugin-elasticsearch/blob/master/lib/fluent/plugin/out_elasticsearch.rb#L291

If `@logstash_prefix_key` should also be able to handle parent/child nesting, then similar logic is needed.

portante:
Working fixes through ViaQ project first.

PR #279, opened 2017-08-23
WIP: Retry on certain errors from Elasticsearch

richm:
Work-In-Progress: more work is needed to verify the emitted logs
are as expected.

There are five cases for an exception is raised to cause the output
buffer thread to retry an operation:

  1. ElasticsearchVersionMismatch - We did not receive an error
     payload that matches our expectations, so this is considered
     a possible Elasticsearch version mismatch
  2. UnrecognizedElasticsearchError - We encountered some error
     from Elasticsearch which we don't recognize
  3. ElasticsearchOutOfMemory - If Elasticsearch is responding, but
     telling us it is out of memory, status code 500, with an error
     type of "out_of_memory_error", we'll keep retrying until it
     returns to us successfully
  4. BulkIndexQueueFull - If Elasticsearch responds with a bulk
     index request rejected error, 429, with an error type of
     "es_rejected_execution_exception", we'll keep retrying until
     it returns to us successfully
  5. ElasticsearchError - Some other status code 500 with an
     unknown error type.

We also now:

  * log at debug level all documents rejected because of bad arguments
  * log at warn level a summary of the number of documents with bad
    arguments
  * log at info level a summary of the number of duplicate documents
    (if the write_operation is "create")
  * log at debug level a summary of all the documents successfully
    indexed

Note we only perform these changes currently for the "static" version of
the plugin.

This commit also fixes the dependencies for Elasticsearch 2.x and
Fluentd v0.12.

DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,10 @@
 source 'https://rubygems.org'
 
 # Specify your gem's dependencies in fluent-plugin-elasticsearch.gemspec
+gem 'fluentd', '~> 0.12.0'
+
 gemspec
+
 
 gem 'simplecov', require: false
 gem 'coveralls', require: false
--- a/Gemfile.v0.14
+++ b/Gemfile.v0.14
@@ -1,10 +1,7 @@
 source 'https://rubygems.org'
 
 # Specify your gem's dependencies in fluent-plugin-elasticsearch.gemspec
-gem 'fluentd', '~> 0.12.0'
-
 gemspec
-
 
 gem 'simplecov', require: false
 gem 'coveralls', require: false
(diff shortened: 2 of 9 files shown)


richm:
@portante I split your original PR into two - this one is about the bulk error handling.  The major difference is that in the default case, we submit a new record like this: `{"index":{"_index":"name","_type":"thetype"}}` and Elasticsearch will convert the `"index"` into a `"create"` in the response, since it technically is a create.  It makes sense, but it is an undocumented behavior of Elasticsearch.

Issue #233, opened 2017-01-06
Add option logstash_prefix_key?


javefang:
#### Problem
I'm using fluentd and the elasticsearch plugin to collect logs from our k8s cluster. My log message have roughly the following format

```
{ "kubernetes": { "namespace_name": "dev" }, "log": "this is a sample log" }
```

For now I'm using `logstash_format true`, which will index the logs into indices `logstash-yyyy.mm.dd`. This is not ideal since the k8s cluster is shared and apps in each namespace write logs with different fields which can potentially cause a mapping conflict.

Ideally I would like to pick the index prefix with field value `kubernetes.namespace_name`. Looking through the documentation I find it's not currently possible. Does it make sense to add an option called `logstash_prefix_key`?

#### Steps to replicate
(None)

bz-keitaro-kobayashi:
We use `fluent-plugin-forest` to work around this limitation, as detailed in the [docs here](https://github.com/uken/fluent-plugin-elasticsearch#not-seeing-a-config-you-need). It would be nice to not use forest for large amounts of tag patterns, though. There's also `elasticsearch_dynamic` in the example below that, with its own concerns.

richm:
https://github.com/uken/fluent-plugin-elasticsearch/pull/281

cosmo0920:
#288 had been shipped.
This feature is available in 2.x versions.
For 1.x versions, I think that this should be a limitation.

PR #281, opened 2017-08-29
Add support for logstash_prefix_key and logstash_prefix_separator

richm:
This allows the use of `@type elasticsearch` instead of dynamic in
many cases where the prefix can be created earlier in the pipeline.

DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -30,6 +30,8 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
   config_param :time_precision, :integer, :default => 0
   config_param :logstash_format, :bool, :default => false
   config_param :logstash_prefix, :string, :default => "logstash"
+  config_param :logstash_prefix_key, :string, :default => nil
+  config_param :logstash_prefix_separator, :string, :default => "-"
   config_param :logstash_dateformat, :string, :default => "%Y.%m.%d"
   config_param :utc_index, :bool, :default => true
   config_param :type_name, :string, :default => "fluentd"
@@ -86,6 +88,10 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
 
     if @remove_keys_on_update && @remove_keys_on_update.is_a?(String)
       @remove_keys_on_update = @remove_keys_on_update.split ','
+    end
+
+    if @logstash_prefix_key && @logstash_prefix_key.is_a?(String)
+      @logstash_prefix_key = @logstash_prefix_key.split '.'
     end
 
     if @template_name && @template_file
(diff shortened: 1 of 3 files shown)


cosmo0920:
I want to add this functionality using with Fluentd v0.14 API ([extract_placeholders](https://docs.fluentd.org/v0.14/articles/api-plugin-output#understanding-chunking-and-metadata)).
Currently, I don't want to add this in `out_elasticsearch`.

richm:
@cosmo0920 I don't understand the question?

cosmo0920:
What does "the question" mean?
Do you want to add dynamic like feature into `elasticsearch`?
If so, I don't want to add this feature into this plugin as-is.

`elasticsearch_dynamic` should be deprecated in the near future, but it should not be until using traditional Fluentd API.

cosmo0920:
Related to https://github.com/uken/fluent-plugin-elasticsearch/issues/233.

richm:
The point of this PR is that it allows us to use `@type elasticsearch` instead of `@type elasticsearch_dynamic`, so we don't need to add it to `elasticsearch_dynamic`.  Are you saying that you would like to have this feature added to `elasticsearch_dynamic` before you will accept the PR?

cosmo0920:
> The point of this PR is that it allows us to use @type elasticsearch instead of @type elasticsearch_dynamic, so we don't need to add it to elasticsearch_dynamic. Are you saying that you would like to have this feature added to elasticsearch_dynamic before you will accept the PR?

No. 
Currently, we can implement this feature with fluent-plugin-forest in Fluentd v0.12.
We should keep plugin simple.

richm:
>  Currently, we can implement this feature with fluent-plugin-forest in Fluentd v0.12.

Well, we would also like to not have to introduce another plugin in order to implement this feature.

cosmo0920:
Again, `elasticsearch` and fluent-plugin-forest can implement this feature.
This is _truly_ Fluentd v0.12 way.
Fluentd plugins should not be monolithic.
In the near future, we can introduce this feature by v0.14 builtin placeholders.
It is alternative and flexible way to implement placeholders in v0.14. 

I'm focused on releasing #223 as 2.0.0.rc1.
I'll come back this after releasing v0.14 adopted rc version.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -88,6 +90,10 @@ def configure(conf)
       @remove_keys_on_update = @remove_keys_on_update.split ','
     end
 
+    if @logstash_prefix_key && @logstash_prefix_key.is_a?(String)
+      @logstash_prefix_key = @logstash_prefix_key.split '.'
```
cosmo0920: I'm planning to support this feature with v0.14's [extract_placeholders](https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L686).
Could you use this method to extract placeholders?
This is more generic way in v0.14/

cosmo0920: #288

cosmo0920:
Closing this in favor of 288.

PR #282, opened 2017-08-29
Retry on certain errors from Elasticsearch

richm:
There are five cases for an exception is raised to cause the output
buffer thread to retry an operation:

  1. ElasticsearchVersionMismatch - We did not receive an error
     payload that matches our expectations, so this is considered
     a possible Elasticsearch version mismatch
  2. UnrecognizedElasticsearchError - We encountered some error
     from Elasticsearch which we don't recognize
  3. ElasticsearchOutOfMemory - If Elasticsearch is responding, but
     telling us it is out of memory, status code 500, with an error
     type of "out_of_memory_error", we'll keep retrying until it
     returns to us successfully
  4. BulkIndexQueueFull - If Elasticsearch responds with a bulk
     index request rejected error, 429, with an error type of
     "es_rejected_execution_exception", we'll keep retrying until
     it returns to us successfully
  5. ElasticsearchError - Some other status code 500 with an
     unknown error type.

We also now:

  * log at debug level all documents rejected because of bad arguments
  * log at warn level a summary of the number of documents with bad
    arguments
  * log at info level a summary of the number of duplicate documents
    (if the write_operation is "create")
  * log at debug level a summary of all the documents successfully
    indexed

Note we only perform these changes currently for the "static" version of
the plugin.

This commit also fixes the dependencies for Elasticsearch 2.x and
Fluentd v0.12.

DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -14,6 +14,11 @@ require_relative 'elasticsearch_index_template'
 
 class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
   class ConnectionFailure < StandardError; end
+  class BulkIndexQueueFull < StandardError; end
+  class ElasticsearchOutOfMemory < StandardError; end
+  class ElasticsearchVersionMismatch < StandardError; end
+  class UnrecognizedElasticsearchError < StandardError; end
+  class ElasticsearchError < StandardError; end
 
   Fluent::Plugin.register_output('elasticsearch', self)
 
@@ -280,8 +285,10 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
     bulk_message = ''
     header = {}
     meta = {}
-
+    records = 0
+    bulk_message_count = 0
     chunk.msgpack_each do |time, record|
+      records += 1
       next unless record.is_a? Hash
 
       if @flatten_hashes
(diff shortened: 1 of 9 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -349,12 +357,95 @@ def get_parent_of(record, path)
...
   end
 
-  def send_bulk(data)
+  def send_bulk(data, count)
     retries = 0
     begin
       response = client.bulk body: data
       if response['errors']
-        log.error "Could not push log to Elasticsearch: #{response}"
+        errors = Hash.new(0)
```
cosmo0920: Any chance to extract into method this _too_ long error classifying implementation?
This implementation is not easy to be testable.

richm: would you like everything in the `if response['errors']` block to be in a separate method, or would you like it broken down more?

cosmo0920: I want to wrap these lines into one or a few encapsulated method.

cosmo0920: Or, extract `HandleError` class.

tschroeder-zendesk:
Can we get some updates on this? This would be extremely helpful for me. I basically have to disable logging at this point because my logs are useless as they are full of conflicts from using a hash to prevent duplicates and so all I ever see is super long lines full of the response which 90% of that is success. Breaking up the bulk requests to the ones that actually fail and reporting those numbers seems much better. At the very least the current log level of error when there is any kind of error in the bulk request and then displaying the entire returned response from ES is way overkill. If I tail -f my log I have to wait about 5 minutes just to cancel it.

richm:
I don't really have time to work on this right now.  We got it to work for our use case: https://github.com/ViaQ/fluent-plugin-elasticsearch/tree/master/lib/fluent/plugin
Feel free to grab that code, rebase to the latest, restructure the code as requested in the PR above, and submit your own PR.

tschroeder-zendesk:
Thanks I'll take a look

cosmo0920:
Closing this in favor of #324, and #322.
Anyway, thank you for your work, @portante and @richm !

PR #288, opened 2017-09-11
Add placeholders support for index_name and logstash_prefix parameters

cosmo0920:
This is v0.14 way to extract placeholders in fluentd conf.

${tag} will be replaced with events's tag:

With `[test.elasticsearch, #<EventTime>, #<Records>]`,
`${tag}` will be replaced into `test.elasticsearch`.

DESCRIPTION HERE

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible for 2.0.0.rc.1 (Note: This patch does not work with Fluentd v0.12)

---

This is another approach of fixing #233.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Current maintainers: @cosmo0920
   + [Hash flattening](#hash-flattening)
   + [Not seeing a config you need?](#not-seeing-a-config-you-need)
   + [Dynamic configuration](#dynamic-configuration)
+  + [Placeholders](#placeholders)
 * [Contact](#contact)
 * [Contributing](#contributing)
 * [Running tests](#running-tests)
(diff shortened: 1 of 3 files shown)


cosmo0920:
@javefang @richm How about this patch?
This patch is not introduced newly parameter.
But placeholders support is introduced against index_name and logstash_prefix parameters.

javefang:
Thank you for the great work @cosmo0920 , do I understand it correctly that this PR will allow the event tag to be used in the index name?

cosmo0920:
Yes. It is basically correct.

v0.14 placeholders can use `${tag}` for tag, `%Y%m%d` like strftime format, and custom record keys like as `record["mykey"]`.

They are used as below:

#### tag

```aconf
<match my.logs>
  @type elasticsearch
  index_name elastic.${tag} #=> replaced with each event's tag. e.g.) elastic.test.tag
  <buffer tag>
    @type memory
  </buffer>
  # <snip>
</match>
```

#### time

```aconf
<match my.logs>
  @type elasticsearch
  index_name elastic.%Y%m%d #=> e.g.) elastic.20170811
  <buffer tag, time>
    @type memory
    timekey 3600
  </buffer>
  # <snip>
</match>
```

#### Custom key

```log
records = {key1: "value1", key2: "value2"}
```

```aconf
<match my.logs>
  @type elasticsearch
  index_name elastic.${key1}.${key2} # => e.g.) elastic.key1.key2
  <buffer tag, key1, key2>
    @type memory
  </buffer>
  # <snip>
</match>
```

cosmo0920:
I'll add v0.14 placeholders documentation for tag, time, and custom key.

richm:
How do placeholders work e.g. 
```
index_name elastic.${key1}.${key2} # => e.g.) elastic.key1.key2
```
Are `${key1}` and `${key2}` evaluated by doing separate ruby `eval`s for each `${....}`, just like in the `out_elasticsearch_dynamic.rb` code?  Because that's why we implemented https://github.com/uken/fluent-plugin-elasticsearch/pull/281 the way we did - those ruby `eval`s were expensive.

Also, I thought, for `record_transformer` anyway, that doing `${keyname}` was bad, and `${record['keyname']}` should be used instead?

cosmo0920:
> Are ${key1} and ${key2} evaluated by doing separate ruby evals for each ${....}, just like in the out_elasticsearch_dynamic.rb code? Because that's why we implemented #281 the way we did - those ruby evals were expensive.

Fluentd v0.14's `extract_placeholders` does not use Ruby eval. It uses `String#gsub` and `Time#strftime` to implement placeholders: https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/output.rb#L686
> Also, I thought, for record_transformer anyway, that doing ${keyname} was bad, and ${record['keyname']} should be used instead?

extract_placeholders behaviour is currently undocumented. But test code says that `${keyname}` is correct: https://github.com/fluent/fluentd/blob/master/test/plugin/test_output.rb#L282

PR #297, opened 2017-09-15
Add logstash_prefix_separator config parameter

### LINKED ISSUE #253: Make target index more flexible
#### Problem
Our company needs an index with the pattern of <name>_<date>, for example `systemd_2017-04-26`. But in your plugin the index pattern is hard-coded to the hyphen `systemd-2017-04-26`.

We now using the workaround with the following configuration that will create an index like `systemd-_2017-04-26`

```
   logstash_dateformat "_%Y.%m.%d"
   logstash_format true
```

Would it be possible to have both the date and to use an underscore instead of a hyphen?

cosmo0920:
Because logstash_prefix and datetime format strings are joined with
hardcoded with "-".
With this parameter, users can use arbitrary characters there.

Fixes #253 for v0.14.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible for v2.0.0.rc.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Current maintainers: @cosmo0920
   + [user, password, path, scheme, ssl_verify](#user-password-path-scheme-ssl_verify)
   + [logstash_format](#logstash_format)
   + [logstash_prefix](#logstash_prefix)
+  + [logstash_prefix_separator](#logstash_prefix_separator)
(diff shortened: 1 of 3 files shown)


cosmo0920:
backported: https://github.com/uken/fluent-plugin-elasticsearch/commit/106803c228df153e496916aaa7011c836ce8538a

Issue #312, opened 2017-10-25
Duplicate events are indexed with different _id field


dannyk81:
#### Problem

We've been experiencing few incidents in the past few weeks where a certain issue on our Elasticsearch cluster would cause errors in fluentd when pushing logs (specifically, we've been exceeding the allowed queue length on Elastic which would send reject errors), these are being retried on fluentd at certain intervals however each retry is submitted with a new `_id` value.

During this condition. Elasticsearch would seem to be indexing the messages (with some delay), but since fluentd is getting an error, it seem to retry again and again, as a result we end up with many duplicate events (each with a unique `_id`)

I've found a related issue filed here: https://bugzilla.redhat.com/show_bug.cgi?id=1491401

I'm not certain if this is the plugin's fault or perhaps the output buffer in fluentd that is responsible for this part.

We were able to come up with the following workaround:

1) We've defined a new filter using the [fluent-plugin-genhashvalue](https://github.com/mtakemi/fluent-plugin-genhashvalue) that generates a `_hash` field from several keys in each record (hopefully, providing a unique md5 hash per each event)
2) Added `id_key _hash` to the Output configuration

The above solved the duplicate issue we've been experiencing.

#### Steps to replicate

This is triggered when Elasticsearch sends rejects due to queue limit reached.

#### Using Fluentd and ES plugin versions

* Fluentd v0.12.35
* ES plugin 1.10.1
* Elasticsearch 5.5.0

tschroeder-zendesk:
This has been a major issue for me as well. I ended up using my a similar workaround, but it would be ideal if this can get fixed.

cosmo0920:
> We've defined a new filter using the fluent-plugin-genhashvalue that generates a _hash field from several keys in each record (hopefully, providing a unique md5 hash per each event)
> Added id_key _hash to the Output configuration

I think that using fluent-plugin-genhashvalue with the following config solves this issue.

```aconf
<filter **>
  @type genhashvalue
  set_key _id
</filter>
```

Fluentd plugins should be used in several combinations.

This issue is caused by heavily traffic congestion in Elasticsearch.
I have no idea to fix this issue in plugin side, but @dannyk81 's workaround is pretty reasonable. 
Could you add your workaround in document?
It might be helpful for heavily traffic congested EFK stack users.

cosmo0920:
> By default, all records inserted into ElasticSearch get a random _id. This option allows to use a field in the record as an identifier.

[...]

PR #318, opened 2017-11-15
Add support to generate hash id for preventing duplicate records

cosmo0920:
refs: #312, https://bugzilla.redhat.com/show_bug.cgi?id=1491401#c15

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920:
@dannyk81, @richm How about this hashed id generatation fix for congested elasticsearch servers?
This PR is naive implementation.
What is your looking for?
Thanks!

dannyk81:
Thanks @cosmo0920!

How about instead of cherry-picking the keys to use for the hash, we instead use the entire record? this should make things unique per record and remove the headache from the user to pick the keys, also (as it is in our case) different sources will require different keys so may not be useable for everyone.

This is indeed how it was suggested [here](https://bugzilla.redhat.com/show_bug.cgi?id=1491401#c15):
```
Hash(record) + Random(8) + Timestamp(usec)
```

portante:
I support the notion of using a sub-set of keys to define the hash, which allows one to define the core record collected at the source, and then decorate it via each handler along the way so that we know we are not handling different records.

portante:
Can we get this to work in both output plugins?

dannyk81:
Hi @portante

Not sure I understand your reasoning, the goal of this hash is to create a unique `_id` for every record being handled by fluentd, this already happens "automatically" today as fluentd generates a totally random `_id` for each record (so there's no way to "reproduce it") and to be honest I don't see a reason why you would need this in this context.

The problem being addressed here is that fluentd will provide a *different* `_id` value for the same record over each retry attempt, which (under certain conditions) causes deplicate events being indexed in Elasticsearch.

The goal here is to create a controlled hash that will persist across retries of the fluentd (as it would be a key of the record itself), thus preventing Elasticsearch from indexing the same event more than once, as such it is critical to ensure that the _hash for every record is unique - this is the main requirement.

The formula suggested in the mentioned Bugzilla seem to meet this requirement as the results consists of the hash of the entire record, a timestamp (in usec) and a random generated sequence.

Using a sub set of keys proves probelmatic in certain conditions, for example - we have multiple sources that have different record structure and different keys, however the hash definition will be the same for all (as it's defined in the output plugin configuration). Today we use a separate filter with the fluent-plugin-genhashvalue plugin to construct the correct _hash for each source, but it's a management headache and we've already had unfortunate hash-collisions, perhaps we can agree on a middle-ground? use the entire record by-default or allow the user to configure the desired subset keys?

richm:
@dannyk81 

> this already happens "automatically" today as fluentd generates a totally random _id for each record

Really?  I thought by default fluent-plugin-elasticsearch would send the records to Elasticsearch with no `_id` field, and let Elasticsearch auto-generate the `_id` value?  If fluentd is doing this, can you point me to that code?

dannyk81:
@richm Sorry!!! I should be more careful when writing long answers before coffee :( I just noticed I've written fluentd in some places where Elasticsearch was applicable (I have updated my original post...)

Indeed, if the `_id` is not provided by fluentd, *Elasticsearch* is the one to generate it, and AFAIU it's quite random:
```
Autogenerated IDs are 20 character long, URL-safe, Base64-encoded GUID strings. These GUIDs are generated from a modified FlakeID scheme which allows multiple nodes to be generating unique IDs in parallel with essentially zero chance of collision.
```
[source](https://www.elastic.co/guide/en/elasticsearch/guide/current/index-doc.html)

portante:
> The goal here is to create a controlled hash that will persist across retries of the fluentd (as it would be a key of the record itself), thus preventing Elasticsearch from indexing the same event more than once, as such it is critical to ensure that the _hash for every record is unique - this is the main requirement.

@dannyk81, I agree, adding an option to allow a simple, Elasticsearch equivalent random ID [3], for those that don't care about repeatable/verifiable IDs is worth adding here.

As a broader discussion topic for all, to consider further the discussion of IDs, can we tell the difference between an ID value generated in these three ways:

 1. `Hash(record + timestamp)`
 2. `Hash(record + timestamp + Random(8))`
 3. `Hash(Random(20))`

As I understand how these would behave, all three have the potential of generating a good unique ID value (assuming `Hash()` follows good ID guidelines [1][2], and `Random(N)` means to generate a random N character string; it should be assumed for this discussion that `Hash(Random(20))` produces the same unique values as the default Elasticsearch algorithm does [3]).

#1 is repeatable and verifiable, meaning one can query Elasticsearch for the `_source` field and re-apply the hash algorithm to arrive at the same `_id` field value.  But it requires having sufficient data in the record that grants uniqueness, as you mention that with [4] you have problems getting the keys right.

#2 is not repeatable or verifiable from the above definition because it includes a random value.

#3 is also not repeatable or verifiable from the above definition because it IS a pure random value.

From inspecting just the `_id` value and having access to the `_source` field, it should be impossible for someone to determine whether #2 or #3 was used to generate the `_id` value.  Also, wouldn't #3 be potentially much faster than #2?

So if we could have an option for performing #3 when we don't have sufficient uniqueness in the keys, we'd have an implementation that implements #1 and #3.

That then begs the question, why not just use [4] when we need IDs derived from fields, and just have a PR that implements #3?

[1] https://www.elastic.co/guide/en/elasticsearch/guide/2.x/indexing-performance.html#_other
[2] http://blog.mikemccandless.com/2014/05/choosing-fast-unique-identifier-uuid.html
[3] https://www.elastic.co/guide/en/elasticsearch/guide/current/index-doc.html
[4] https://github.com/mtakemi/fluent-plugin-genhashvalue

cosmo0920:
I implemented the following in this PR.

```
Hash(sub set of record value) + Random(8) + (nanosecods in timestamp)
```

I agree to change the following implementation:

```ruby
# Hash(Random(8)
record[@hash_config.hash_id_key] = Digest::SHA256(SecureRandom.hex(20))
```

And I noticed that we can use Fluentd's chunk_id mechanism for generating hash id.
`chunk.unique_id` is unique entire in chunk.
Thus, another implementation is:

```ruby
record[@hash_config.hash_id_key]  = Digest::SHA256(dump_unique_id_hex(chunk.unique_id))
```

Which is better to implement this feature?

dannyk81:
@portante 

> That then begs the question, why not just use [4] when we need IDs derived from fields, and just have a PR that implements #3?

I believe this is a solid approach, since this is aligned with the way Elasticsearch generates the `_id` it makes total sense to pick up that approach here, and indeed if another option is needed, [4] can be used.

cosmo0920:
I updated this PR implementation.
I implemented `#4`.

README.md
```diff
@@ -552,6 +553,23 @@ This will produce elasticsearch output that looks like this:
...
+
+Fluentd re-emits events that failed to be indexed/ingested in Elasticsearch with a new and unique `_id` value, this means that congested Elasticsearch clusters that reject events (due to command queue overflow, for example) will cause Fluentd to re-emit the event with a new `_id`.
+
+Using generated hash id is another fix way to prevent duplicate `_id` in elasticsearch records.
+Here is a sample config:
+
+```
+<hash>
+  hash_type sha256   # md5, sha1, sha256, sha512 are available here.
+  hash_id_key _id    # storing genrated hash id key
```
portante: s/genrated/generated/

README.md
```diff
@@ -552,6 +553,23 @@ This will produce elasticsearch output that looks like this:
 
 Note that the flattener does not deal with arrays at this time.
 
+### Generate Hash ID
+
+Fluentd re-emits events that failed to be indexed/ingested in Elasticsearch with a new and unique `_id` value, this means that congested Elasticsearch clusters that reject events (due to command queue overflow, for example) will cause Fluentd to re-emit the event with a new `_id`.
```
portante: Perhaps consider rewording this line to read, "By default, the fluentd elasticsearch plugin does not emit records with a `_id` field, leaving it to Elasticsearch to generate a unique `_id` as the record is indexed.  When an Elasticsearch cluster is congested and begins to take longer to respond than the configured `request_timeout`, the fluentd elasticsearch plugin will re-send the same bulk request.  Since Elasticsearch can't tell its actually the same request, all documents in the request are indexed again resulting in duplicate data.  In certain scenarios, this can result in essentially and infinite loop generating multiple copies of the same data."

README.md
```diff
@@ -552,6 +553,23 @@ This will produce elasticsearch output that looks like this:
...
+
+Using generated hash id is another fix way to prevent duplicate `_id` in elasticsearch records.
```
portante: Perhaps consider rewording this as well to read, "Using an `_id` generated by the fluentd elasticsearch plugin will communicate to Elasticsearch the uniqueness of the requests so that duplicates will be rejected or simply replace the existing records."

lib/fluent/plugin/generate_hash_id_support.rb
```diff
@@ -0,0 +1,36 @@
...
+        when :sha256
+          Digest::SHA256.hexdigest(str)
+        when :sha512
+          Digest::SHA512.hexdigest(str)
+        end
+      end
+
+      def generate_hash_id_key(record)
+        s = ""
+        s += hash_hex(SecureRandom.hex(20))
```
portante: Does this method adhere to the recommendations of good lucene keys as described at [1]?

[1] https://www.elastic.co/guide/en/elasticsearch/guide/current/index-doc.html

README.md
```diff
@@ -552,6 +553,23 @@ This will produce elasticsearch output that looks like this:
...
+
+```
+<hash>
+  hash_type sha256   # md5, sha1, sha256, sha512 are available here.
+  hash_id_key _id    # storing genrated hash id key
+<hash>
+id_key _hash_id
+```
+
+:warning: In order to avoid hash-collisions and loosing data careful consideration is required when choosing the keys in the event record that should be used to calculate the hash.
```
portante: Seems like with this update you can remove this warning.

README.md
```diff
@@ -552,6 +553,22 @@ This will produce elasticsearch output that looks like this:
...
+Here is a sample config:
+
+```
+<hash>
+  hash_type sha256   # md5, sha1, sha256, sha512 are available here.
+  hash_id_key _id    # storing genrated hash id key
+<hash>
+```
+
+:warning: In order to avoid hash-collisions and loosing data careful consideration is required when choosing the keys in the event record that should be used to calculate the hash.
```
portante: Seems like this warning can be removed now.

lib/fluent/plugin/generate_hash_id_support.rb
cosmo0920: Elasticsearch document says that:

> Autogenerated IDs are 20 character long, URL-safe, Base64-encoded GUID strings. 

I think that we can implement similar functionality as below:

```ruby
s += Base64.encode64(SecureRandom.uuid)
```

But this line generates 48 characters, though.

README.md
dannyk81: 👍

lib/fluent/plugin/generate_hash_id_support.rb
cosmo0920: I've used this implementation for now.

cosmo0920:
If there is no comments or request changes within a few days, I'll release these patches as 2.0.1.rc.1.

lukas-vlcek:
I am confused. The proposed update to doc reads:

> :warning: In order to avoid hash-collisions and loosing data careful consideration is required when choosing the keys in the event record that should be used to calculate the hash

However, in the code I see:

```ruby
s += Base64.strict_encode64(SecureRandom.uuid)
record[@hash_config.hash_id_key] = s
```

**Questions one:** Is user able to define which keys/records are used to calculate the hash? (As the docs suggest) Or the hash is always `SecureRandom.uuid` and user has no way to change that? (In which case the docs needs update).

**Questions two:** What is the persistence story for the generated hash value? If fluentd is restarted while sending the data is there a chance that the same record will be picked by fluentd again and send to ES with different `_id` value? (This goes back to `_id` reproducibility, ie what @portante was talking about).

cosmo0920:
You are reading old version of proposed documents?
Not related to change in #317.

> Questions one: Is user able to define which keys/records are used to calculate the hash? (As the docs suggest) Or the hash is always SecureRandom.uuid and user has no way to change that? (In which case the docs needs update).

No way to change by users in current patches.

> Questions two: What is the persistence story for the generated hash value? If fluentd is restarted while sending the data is there a chance that the same record will be picked by fluentd again and send to ES with different _id value? (This goes back to _id reproducibility, ie what @portante was talking about).

There is no way to ensure _id reproducibility.
`SecureRandom.uuid` ensure only generated hash uniqueness.
Using file buffer to ensure sending records even if buffer is remained at reboot fluentd instance.
But uuid collision will occur:

2^(122/2)=2305843009213693952≈2.3 ∗ 10^18

after generating times.

lukas-vlcek:
@cosmo0920 I believe I am reading current patch docs proposal, see https://github.com/uken/fluent-plugin-elasticsearch/pull/318/files#diff-04c6e90faac2675aa89e2176d2eec7d8R413

lukas-vlcek:
@cosmo0920 

> Ah! This line just removes a trailing space.

But the content is still part of PR, no?

> There is no way to ensure _id reproducibility.

I think there are various options how to look at this. For example, we can think of a one simple fluentd collecting logs, adding `_flunetd_id` key with random value and sending the record to another fluentd. This another fluentd will use `_source` to calculate unique yet reproducible `_id` from all or selected keys including the `_flunetd_id`. I think this is strategy that can make sense in practice (and is flexible IMO).

I am not saying we should implement this in this PR.

cosmo0920:
>> Ah! This line just removes a trailing space.

> But the content is still part of PR, no?

No. I removed this line in the PR patch.

> I think there are various options how to look at this. For example, we can think of a one simple fluentd collecting logs, adding _flunetd_id key with random value and sending the record to another fluentd. This another fluentd will use _source to calculate unique yet reproducible _id from all or selected keys including the _flunetd_id. I think this is strategy that can make sense in practice (and is flexible IMO).

I'm sorry I didn't understand what you mean.
Both of nodes' Fluentd will be down-able.
It seems to be just increasing system complexity.

lukas-vlcek:
> Both of nodes' Fluentd will be down-able.
> It seems to be just increasing system complexity.

The point is that for the second fluentd node the calculation of _id is perfectly reproducible (no need to calculate random value) and the first node can be replaced with something different in the future. For example the data itself changes and random id value becomes part of it.

cosmo0920:
> The point is that for the second fluentd node the calculation of _id is perfectly reproducible (no need to calculate random value)

If you want to make hash calculation to be reproducible, you should create some output fluentd plugin which stores hash and some record related unique id and calculate hash.

Using fluent-plugin-genhashvalue is suitable for your scenario.
But fluent-plugin-genhashvalue is still using old Fluentd's filter plugin API.
I'm trying to migrate to use v0.14 API in [this PR](https://github.com/mtakemi/fluent-plugin-genhashvalue/pull/1).
This is an obstacle for using multi workers feature in Fluentd v0.14.

PR #356, opened 2018-02-05
Fix timestamp precision issue when using time_key and time_key_format

meetyg:
This is an old issue that hasn't been fixed...

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -376,7 +376,7 @@ module Fluent::Plugin
           elsif record.has_key?(@time_key)
             rts = record[@time_key]
             dt = parse_time(rts, time, tag)
-            record[TIMESTAMP_FIELD] = rts unless @time_key_exclude_timestamp
+            record[TIMESTAMP_FIELD] = dt.iso8601(@time_precision) unless @time_key_exclude_timestamp
           else
             dt = Time.at(time).to_datetime
             record[TIMESTAMP_FIELD] = dt.iso8601(@time_precision)
--- a/test/plugin/test_out_elasticsearch.rb
+++ b/test/plugin/test_out_elasticsearch.rb
@@ -1265,7 +1265,7 @@ class ElasticsearchOutput < Test::Unit::TestCase
     driver.configure("logstash_format true
                       time_key_format %Y-%m-%d %H:%M:%S.%N%z
                       time_key vtm
-                      time_precision 0\n")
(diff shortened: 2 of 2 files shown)


cosmo0920:
Could you fix failing test case?

meetyg:
Yes, but for some odd reason, it seems that @time_precision is not 0 in tests.
Do you have an idea ?

cosmo0920:
What is the relevant issue?

meetyg:
Actually, I want to validate my understanding of the proper expected behavior:
When using `logstash_format true`, this means that the `@timestamp` field should be in the format elasticsearch/kibana expects, meaning: `%Y-%m-%dT%H:%M:%S.%NZ` or `2016-02-16T16:04:05.930-08:00` for example.

If so, the unit tests need to updated, so that when `logstash_format true` or `include_timestamp true` is set, the output `@timestamp` value and expected value **are both in the same format**.

Currently this is not the case in all relevant tests. Most of them compare the values using default formats.

If my understanding is correct, I can update the tests accordingly.

Please verify.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -376,7 +376,7 @@ def write(chunk)
           elsif record.has_key?(@time_key)
             rts = record[@time_key]
             dt = parse_time(rts, time, tag)
-            record[TIMESTAMP_FIELD] = rts unless @time_key_exclude_timestamp
```
cosmo0920: IMO, specifying `time_key` case should handle timestamp as-is instead of iso8601 conversion.

meetyg: I think that it is expected that the plugin be able to parse timestamps in different formats, and output them in the expected "logstash" format.
Not all logs have their timestamps in the logstash format, and to my understanding thats why time_key_format is for.
Just pushing back the unparsed time field value into `@timestamp` field seems awkward.
This can also cause milliseconds to be lost when inserting the  `@timestamp` field into elasticsearch.

I saw your PR #358 , but still you assume that the input is allready in `iso8601` (elastic) format.
Please reconsider this.

cosmo0920: Hmm, the above article's claim is reasonable for me.
Thank you for your suggestion. :+1:

cosmo0920:
I've released #360 as v2.6.0.
Could you check v2.6.0?

PR #371, opened 2018-03-02
Check es major version

cosmo0920:
Currently, ES 6.x cannot handle documents without `_type`.
But, Elasticsearch 6.0 supports a single mapping type per index only.
And referring https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/712,
current ES accepts single fixed type name in `_type` field.

We should follow this logstash-output-elasticsearch behavior. 

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -482,6 +482,7 @@ EOC
         if retries < 2
           retries += 1
           @_es = nil
+          @_es_info = nil
           log.warn "Could not push logs to Elasticsearch, resetting connection and trying again. #{e.message}"
           sleep 2**retries
           retry
@@ -489,6 +490,7 @@ EOC
         raise ConnectionFailure, "Could not push logs to Elasticsearch after #{retries} retries. #{e.message}"
       rescue Exception
         @_es = nil if @reconnect_on_error
+        @_es_info = nil if @reconnect_on_error
         raise
       end
     end
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -414,11 +414,12 @@ EOC
         end
 
         target_type_parent, target_type_child_key = @target_type_key ? get_parent_of(record, @target_type_key) : nil
-        if @last_seen_major_version >= 6
-          log.warn "Detected a 6.x and above: `@type_name` will be used as the document `_type`."
-        end
         if target_type_parent && target_type_parent[target_type_child_key]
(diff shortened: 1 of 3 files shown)


cosmo0920:
How about this change, @sandstrom?

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -33,7 +34,10 @@ class ConnectionFailure < StandardError; end
     config_param :scheme, :string, :default => 'http'
     config_param :hosts, :string, :default => nil
     config_param :target_index_key, :string, :default => nil
-    config_param :target_type_key, :string, :default => nil
+    config_param :target_type_key, :string, :default => nil,
+                 :deprecated => <<EOC
+Elasticsearch 7.x or above will ignore this config. Please use fixed type_name instead.
+EOC
```
sandstrom: How about using this ruby 2.3 functionality to remove the leading whitespace? (depends on your min ruby version though)

```ruby
def test
  <<~END
    First content line.
      Two spaces here.
    No space here.
  END
end

test
# => "First content line.\n  Two spaces here.\nNo space here.\n"
```

Source:
https://stackoverflow.com/a/34888238/118007

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -404,6 +416,10 @@ def write(chunk)
         target_type_parent, target_type_child_key = @target_type_key ? get_parent_of(record, @target_type_key) : nil
         if target_type_parent && target_type_parent[target_type_child_key]
           target_type = target_type_parent.delete(target_type_child_key)
+          if @last_seen_major_version >= 6
+            log.warn "Detected ES 6.x and above: `@type_name` will be used as the document `_type`."
+            target_type = type_name
+          end
```
sandstrom: How about adding code for 7 too?

```ruby
if @last_seen_major_version >= 7
  target_type = '_doc'
end
```

lib/fluent/plugin/out_elasticsearch.rb
cosmo0920: I wish I could.
But this plugin should support Ruby 2.1 and Ruby 2.2.
I want to leave as-is this nor now. :confounded:

lib/fluent/plugin/out_elasticsearch.rb
cosmo0920: I've fixed it.

cosmo0920:
I'll release these changes as v2.8.0 and I won't backport them to 1.x.

PR #387, opened 2018-03-23
Use nanosecond precision

martinssipenko:
This change adds ability to specify `time_precision` when using `elasticsearch_dynamic`.

Adding sub-second precision does change the default `@timestamp` value, thus might be considered non backwards compatible change, but given the value should be accepted by elastic I think it should be fine.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/test/plugin/test_out_elasticsearch_dynamic.rb
+++ b/test/plugin/test_out_elasticsearch_dynamic.rb
@@ -479,13 +479,33 @@ class ElasticsearchOutputDynamic < Test::Unit::TestCase
     driver.configure("logstash_format true\n")
     stub_elastic_ping
     stub_elastic
-    ts = DateTime.now
-    time = Fluent::EventTime.from_time(ts.to_time)
+    begin
+      time = Fluent::EventTime.new(Time.now.to_i, 000000000)
+    rescue
+      time = Fluent::Engine.now
+    end
     driver.run(default_tag: 'test') do
-      driver.feed(time, sample_record)
+      driver.feed(time.to_i, sample_record)
     end
     assert(index_cmds[1].has_key? '@timestamp')
-    assert_equal(index_cmds[1]['@timestamp'], ts.iso8601)
+    assert_equal(index_cmds[1]['@timestamp'], Time.at(time).iso8601(9))
+  end
+
+  def test_uses_subsecond_precision_when_configured
+    driver.configure("logstash_format true
+                      time_precision 3\n")
+    stub_elastic_ping
+    stub_elastic
+    begin
+    time = Fluent::EventTime.new(Time.now.to_i, 000000000)
+    rescue
+      time = Fluent::Engine.now
+    end
+    driver.run(default_tag: 'test') do
+      driver.feed(time.to_i, sample_record)
+    end
+    assert(index_cmds[1].has_key? '@timestamp')
+    assert_equal(index_cmds[1]['@timestamp'], Time.at(time).iso8601(3))
   end
 
   def test_uses_custom_timestamp_when_included_in_record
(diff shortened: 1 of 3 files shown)


cosmo0920:
Hmm, elasticsearch-ruby v6.0.2 causes test failure.
I’ll follow up testcases failure.

martinssipenko:
@cosmo0920 thanks, i can rebase if you fix tests in master.

test/plugin/test_out_elasticsearch_dynamic.rb
```diff
@@ -474,13 +474,33 @@ def test_adds_logstash_timestamp_when_configured
     driver.configure("logstash_format true\n")
     stub_elastic_ping
     stub_elastic
-    ts = DateTime.now
-    time = Fluent::EventTime.from_time(ts.to_time)
+    begin
+      time = Fluent::EventTime.new(Time.now.to_i, 000000000)
```
cosmo0920: This plugin already depends on Fluentd v0.14 or later.
`Fluent::EventTime.new` always exists.

test/plugin/test_out_elasticsearch_dynamic.rb
```diff
@@ -474,13 +474,33 @@ def test_adds_logstash_timestamp_when_configured
...
+    rescue
+      time = Fluent::Engine.now
+    end
     driver.run(default_tag: 'test') do
-      driver.feed(time, sample_record)
+      driver.feed(time.to_i, sample_record)
```
cosmo0920: I want to `Fluent::EVentTime` instance value here if possible.

test/plugin/test_out_elasticsearch_dynamic.rb
```diff
@@ -474,13 +474,33 @@ def test_adds_logstash_timestamp_when_configured
...
+                      time_precision 3\n")
+    stub_elastic_ping
+    stub_elastic
+    begin
+    time = Fluent::EventTime.new(Time.now.to_i, 000000000)
+    rescue
+      time = Fluent::Engine.now
+    end
+    driver.run(default_tag: 'test') do
+      driver.feed(time.to_i, sample_record)
```
cosmo0920: I want to use non-integer casted value.
Because v0.14’s `#feed` can accept `Integer` and `Fluent::EventTime` value.

cosmo0920:
I've merged fixing failing testcases PR.
Could you rebase off master?

martinssipenko:
Thanks

cosmo0920:
I've merged PR.
Thank you for your contributions, @martinssipenko !

---

BTW, this is off topic.
Feel free to ignore this.
Why do you use `elasticsearch_dynamic`?
Because I want to almost replace `elasticsearch_dynamic` with Fluentd v0.14 built-in placeholders feature.

martinssipenko:
@cosmo0920 I use dynamic to push to difeerent indices. Eavh record has an app name field in it which i use in prefix. Any better way you can suggest to do this?

cosmo0920:
> @cosmo0920 I use dynamic to push to difeerent indices. Eavh record has an app name field in it which i use in prefix. Any better way you can suggest to do this?

ES plugin 2.x series can handle built-in placeholders in logstash_prefix, index_name, and type_name parameters.
I guess your use case can replace with placeholders in ES plugin like as:

```aconf
<match **>
  @type elasticsearch
  # snip
  <buffer tag, YOUR_RECORD_FIELD> # $.nested.field.can.be.accessible
    @type memory # or file
    # snip
  </buffer>
  index_name app-${YOUR_RECORD_FIELD}
</match>
```
ref: https://github.com/uken/fluent-plugin-elasticsearch#placeholders
ref: https://docs.fluentd.org/v1.0/articles/buffer-section#other-keys

martinssipenko:
@cosmo0920 thanks for suggestion I wasn't aware of such feature. I've done the change on my end and am no more using the dynamic plugin.

Issue #307, opened 2017-10-16
missing authentication token for REST request [/_bulk]


Bobi-zhou:
#### Problem

After running for a while，fluentd warn these logs：

  2017-10-16 08:31:33 +0800 [warn]: suppressed same stacktrace
2017-10-16 08:31:37 +0800 [warn]: retry succeeded. plugin_id="object:3ff1ac0425d0"
2017-10-16 08:32:17 +0800 [warn]: temporarily failed to flush the buffer. next_retry=2017-10-16 08:32:18 +0800 error_class="Elasticsearch::Transport::Transport::Errors::Unauthorized" error="[401] {\"error\":{\"root_cause\":[{\"type\":\"security_exception\",\"reason\":\"missing authentication token for REST request [/_bulk]\",\"header\":{\"WWW-Authenticate\":\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\"}}],\"type\":\"security_exception\",\"reason\":\"missing authentication token for REST request [/_bulk]\",\"header\":{\"WWW-Authenticate\":\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\"}},\"status\":401}" plugin_id="object:3ff1ac0425d0"
  2017-10-16 08:32:17 +0800 [warn]: suppressed same stacktrace
2017-10-16 08:32:18 +0800 [warn]: temporarily failed to flush the buffer. next_retry=2017-10-16 08:32:21 +0800 error_class="Elasticsearch::Transport::Transport::Errors::Unauthorized" error="[401] {\"error\":{\"root_cause\":[{\"type\":\"security_exception\",\"reason\":\"missing authentication token for REST request [/_bulk]\",\"header\":{\"WWW-Authenticate\":\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\"}}],\"type\":\"security_exception\",\"reason\":\"missing authentication token for REST request [/_bulk]\",\"header\":{\"WWW-Authenticate\":\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\"}},\"status\":401}" plugin_id="object:3ff1ac0425d0"
  2017-10-16 08:32:18 +0800 [warn]: suppressed same stacktrace
[...]

Issue #277, opened 2017-08-09
Simple Auth not working with ES5


krlng:
#### Problem
I want to send logs to an AWS ES5.3 through a simple-auth service running in Kubernetes. 

Inserting values using the http type works, but the elasticsearch fluent plugin does neither insert values nor prints any errors. The problem seems to be the basic auth, since it works if I open ES.

#### Steps to replicate

I am using the versions I found [here](https://github.com/uken/fluent-plugin-elasticsearch/issues/231#issuecomment-306039688):

fluentd-plugin-elasticsearch v1.9.5
elasticsearch v5.0.4
elasticsearch-api v5.0.4
elasticsearch-transport v5.0.4

My fluentd configuration looks like    

```
  <source>
      @type tail
      path /var/log/containers/*.log
      pos_file /tmp/es-containers.log.pos
      time_format %Y-%m-%dT%H:%M:%S.%NZ
      tag kubernetes.*
      format json
      read_from_head true
      time_key date
    </source>

    <match kubernetes.**>
      @type copy
      <store>
        @type file
        path /tmp/log/fluent/myapp1
      </store>
      <store>
        @type elasticsearch
        host  "user:pw@my-url"
        scheme https
        index_name kubernetes
        type_name fluentd
      </store>
      <store>
          @type http
          endpoint_url    "#{ENV['ES_HOST']}"
          http_method     post
          serializer      json
          authentication  basic
          username        "#{ENV['ES_USERNAME']}"
          password        "#{ENV['ES_PASSWORD']}"
      </store>
    </match>
```

bechtoldt:
can you show the HTTP request from your client (fluentd) and the HTTP response from your reverse proxy?

bechtoldt:
is this related to #257

cosmo0920:
Is this related to https://github.com/elastic/elasticsearch-ruby/issues/464#issuecomment-356500993?

Issue #257, opened 2017-05-04
401 not authorized


marcofranssen:
Please see the config I'm using to authenticate. For some bizarre reason it just is not able to login.

I also logged the issue over here https://github.com/fluent/fluentd-kubernetes-daemonset/issues/15.

looking forward to a solution.

```
2017-05-03T15:08:36.192683683Z 2017-05-03 15:08:36 +0000 [warn]: temporarily failed to flush the buffer. next_retry=2017-05-03 15:09:06 +0000 error_class="Elasticsearch::Transport::Transport::Errors::Unauthorized" error="[401] " plugin_id="object:2b1bd5d5c250"
```

I do have the following config in place

```fluentd.yaml
---
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  name: fluentd
  namespace: kube-system
  labels:
    k8s-app: fluentd-logging
spec:
  template:
    metadata:
      labels:
        k8s-app: fluentd-logging
    spec:
      containers:
      - name: fluentd
        image: fluent/fluentd-kubernetes-daemonset:v0.12-elasticsearch
        resources:
          requests:
            cpu: 100m
            memory: 200Mi
          limits:
            memory: 200Mi
        env:
        - name: FLUENT_ELASTICSEARCH_HOST
          value: "elasticsearch"
        - name: FLUENT_ELASTICSEARCH_PORT
          value: "9200"
        - name: FLUENT_ELASTICSEARCH_USER
          value: "elastic"
        - name: FLUENT_ELASTICSEARCH_PASSWORD
          value: "changeme"
        volumeMounts:
        - name: fluentconfig
          mountPath: /home/fluent/fluentd/etc
        - name: varlog
          mountPath: /var/log
        - name: varlibdockercontainers
          mountPath: /var/lib/docker/containers
          readOnly: true
      terminationGracePeriodSeconds: 30
      volumes:
      - name: varlog
        hostPath:
          path: /var/log
      - name: varlibdockercontainers
        hostPath:
          path: /var/lib/docker/containers
      - name: fluentconfig
        configMap:
          name: fluent-config
```

In order to handle the `FLUENT_ELASTICSEARCH_USER` and `FLUENT_ELASTICSEARCH_PASSWORD` I have added the following configmap.

```fluentconfig.yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: fluent-config
  namespace: kube-system
data:
  fluent.conf: |
[...]

PR #394, opened 2018-04-09
auth: Fix missing auth tokens after reloading connections

fujimotos:
This patch fixes the "401 not authorized" issue reported by #257, #277 and #307.

### Problem

Essentially, elasticsearch-ruby treats the array of ES nodes (`hosts`) as something
ephemeral and try to reconstruct it periodically using the latest information fetched
from the target cluster.

Since above, as to the settings which should be persistent across the refresh, clients
are supposed to pass them through the `options` parameter, not the `hosts` parameter.

However, fluent-plugin-elasticsearch pass auth tokens as a part of the `hosts` parameter.
Therefore these bits are not retained across a refresh of connections.

### Solution

This patch fixes the issue by making the auth token passed to the client though `options`.

### Checksheet

- [ ] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,19 +106,12 @@ This plugin creates Elasticsearch indices by merely writing to them. Consider us
 
 ```
 hosts host1:port1,host2:port2,host3:port3
-# or
-hosts https://customhost.com:443/path,https://username:password@host-failover.com:443
 ```
 
 You can specify multiple Elasticsearch hosts with separator ",".
 
 If you specify multiple hosts, this plugin will load balance updates to Elasticsearch. This is an [elasticsearch-ruby](https://github.com/elasticsearch/elasticsearch-ruby) feature, the default strategy is round-robin.
 
-And this plugin will escape required URL encoded characters within `%{}` placeholders.
-
-```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -131,7 +124,7 @@ path /elastic_search/
 scheme https
 ```
 
-You can specify user and password for HTTP basic auth. If used in conjunction with a hosts list, then these options will be used by default i.e. if you do not provide any of these options within the hosts listed.
+You can specify user and password for HTTP Basic authentication.
```
cosmo0920: Previous implementation assumes that users can use multiple Elasticsearch cluster in one ES plugin instance.
I reckon that we should write configuration migration guide.

Otherwise, I agree with removing basic authentication description for multiple hosts.

fujimotos: In 4ae5697, I added a small note to tell users to migrate from URL-embedded user credentials.

cosmo0920:
@fujimotos Any chance to backport this change into v0.12 branch?
If you have no enough time to do, I'll send backport PR to v0.12 branch.

fujimotos:
@cosmo0920 Thank you for backporting  the patch to v0.12!

PR #396, opened 2018-04-10
Make the retry_on_failure setting configurable

portante:
We really don't want to retry 5 times on failure.  This can result in 5x the number of bulk requests to Elasticsearch, causing the bulk queues to overflow when you have enough clients.  By making this configurable, folks can correct this behavior by setting the value to zero (0).

Backport of #395.

DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [x] backward compatible


portante:
See #395, not an issue.  Closing.

PR #398, opened 2018-04-10
introduce dead letter queue to handle issues unpacking file buffer chunks

jcantrill:
DESCRIPTION HERE
There are [cases](https://bugzilla.redhat.com/show_bug.cgi?id=1562004) where file buffer chunks become corrupt and block fluentd from processing either good messages or good chunks.  This PR introduces a 'dead letter queue' to drop those messages from the chunk or write them to an 'dlq' file.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Note: For Amazon Elasticsearch Service please consider using [fluent-plugin-aws-
   + [time_parse_error_tag](#time_parse_error_tag)
   + [reconnect_on_error](#reconnect_on_error)
   + [with_transporter_log](#with_transporter_log)
+  + [dlq_handler](#dlq_handler)
   + [Client/host certificate options](#clienthost-certificate-options)
   + [Proxy Support](#proxy-support)
   + [Buffered output options](#buffered-output-options)
@@ -472,6 +473,17 @@ We recommend to set this true if you start to debug this plugin.
 
 ```
 with_transporter_log true
+```
+
+### dlq_handler
+Adds an error handler for processing corrupt messages from message buffers.
+There are [known cases](https://bugzilla.redhat.com/show_bug.cgi?id=1562004) where
+fluentd is stuck processing messages because the file buffer is corrupt.  Fluentd
+is unable to clear faulty buffer chunks.
+
+```
+dlq_handler {'type':'drop'}        #default is to log and drop messages
+dlq_handler {'type':'file', 'dir':'/tmp/fluentd/dlq', 'max_files':5, 'max_file_size':104857600}
 ```
 
 ### Client/host certificate options
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -69,6 +69,7 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOutput
   config_param :reconnect_on_error, :bool, :default => false
   config_param :pipeline, :string, :default => nil
   config_param :with_transporter_log, :bool, :default => false
+  config_param :dlq_handler, :hash, :default => { 'type' =>'drop' }
 
   include Fluent::ElasticsearchIndexTemplate
   include Fluent::ElasticsearchConstants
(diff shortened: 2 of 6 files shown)


jcantrill:
cc @portante @ewolinetz @richm

portante:
@cosmo0920, do you have a sense on how to do that in the fluentd way?  It certainly seems that is a good idea, but we need to be able to communicate which parts of the chunk are good and bad.

For example, there might be 100 records in a chunk, and 13 of them are bad.  We want to have the ES plugin, or any other output plugin, send off the 87 that are good, but return the 13 that are bad.

README.md
```diff
@@ -474,6 +475,17 @@ We recommend to set this true if you start to debug this plugin.
 with_transporter_log true
 ```
 
+### dlq_handler
+Adds an error handler for processing unpacked messages from message buffers.
```
richm: s/unpacked/corrupted/

cosmo0920:
@portante There is no standard Fluentd way to handle good-and-bad contaminated chunk.
I think that this use case is similar to https://github.com/fluent/fluentd/issues/1911.
Umm..., current Fluentd retry mechanism implementation is always retrying until reaching retry limits.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -319,74 +357,77 @@ def write_objects(tag, chunk)
...
-      end
+        if @remove_keys
+            @remove_keys.each { |key| record.delete(key) }
+        end
 
-      if @remove_keys
-        @remove_keys.each { |key| record.delete(key) }
+        append_record_to_messages(@write_operation, meta, header, record, bulk_message)
+        @error.bulk_message_count += 1
+      rescue=>e
```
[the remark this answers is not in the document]
richm: or maybe a `git diff -b` or `git diff -w`

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -129,6 +130,43 @@ def configure(conf)
       log_level = conf['@log_level'] || conf['log_level']
       log.warn "Consider to specify log_level with @log_level." unless log_level
     end
+
+    if @dlq_handler && @dlq_handler.is_a?(Hash)
+      dlq_type = @dlq_handler[:type]||@dlq_handler['type']
+      if dlq_type
```
cosmo0920: Could you use early return style for readability?

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -129,6 +130,43 @@ def configure(conf)
       log_level = conf['@log_level'] || conf['log_level']
       log.warn "Consider to specify log_level with @log_level." unless log_level
     end
+
+    if @dlq_handler && @dlq_handler.is_a?(Hash)
```
cosmo0920: These added lines are bit of long to write in `#configure`.
Could you extract as a method like as `configure_dlq_handler`?

lib/fluent/plugin/out_elasticsearch.rb
jcantrill: I will extract to its own configure method.  Early return doesn't make sense in the current impl given future changes would require configuration to occur before this block; early return could affect how someone else adds

jcantrill:
@cosmo0920 I agree this would be better handled for all output plugins but fluentd does not have a proper way at this time to handle issues associated with processing individual messages from corrupt chunks.  This is a companion issue to not being able to properly handle partial success of processing chunks.  For example, bulk inserts to Elasticsearch where some inserts succeed and others do not.  We shouldn't be retrying the entire insert, just those that failed.

cosmo0920:
> This is a companion issue to not being able to properly handle partial success of processing chunks. For example, bulk inserts to Elasticsearch where some inserts succeed and others do not. We shouldn't be retrying the entire insert, just those that failed.

I see. Current your patches drop entire chunk which is included corrupted or invalid encoding messages.
And I'm re-thinking that we may merge these patches into v0.12 branch because old stable version of Fluentd, which is v0.12, is not actively maintained and not added new features.

Current Fluentd's retrying chunk mechanism does not assume corrupted chunks and v0.12 is not actively maintained.
So, this PR is *reasonable* for me.
But these patches uplift into v1/v0.14 is *not reasonable* for me.
[Backup corrupted chunk feature](https://github.com/fluent/fluentd/issues/1856) is planned for Fluentd v1.

jcantrill:
@cosmo0920 I have some ideas where we could fundamentally introduce this same logic to output.rb.  Move the configuration into lib/fluentd/output.rb and then pass the error handler to be consumed by specific implementations.  Ill work on a pr for that

cosmo0920:
@jcantrill Until then, should we merge and release this PR as v1.14.0?
Fluentd upstream feature request is targeted to v1, *not v0.12*.

cosmo0920:
I've released this PR as v1.14.0.

cosmo0920:
I think that current ES plugin does not handle exceptions.
But with the following patch can handle exception and start to use `@ERROR` label on error:

### For v1

```diff
diff --git a/lib/fluent/plugin/out_elasticsearch.rb b/lib/fluent/plugin/out_elas
ticsearch.rb
index 55c8592..23b7faa 100644
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -375,6 +375,7 @@ EOC
     end
 
     def write(chunk)
+      begin
       bulk_message = ''
       header = {}
       meta = {}
@@ -468,6 +469,9 @@ EOC
         append_record_to_messages(@write_operation, meta, header, record, bulk_
message)
         @error.bulk_message_count += 1
       end
+      rescue => e
+        router.emit_error_event(tag, time, record, e)
+      end
 
       send_bulk(bulk_message) unless bulk_message.empty?
```

### For v0.12 branch

```diff
diff --git a/lib/fluent/plugin/out_elasticsearch.rb b/lib/fluent/plugin/out_elas
ticsearch.rb
index 03f1bcb..38d37ef 100644
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -362,8 +362,8 @@ class Fluent::ElasticsearchOutput < Fluent::ObjectBufferedOu
tput
       next unless record.is_a? Hash
       begin
         process_message(tag, meta, header, time, record, bulk_message)
-      rescue=>e
-        handle_chunk_error(self, tag, e, time, record)
+      rescue => e
+        router.emit_error_event(tag, time, record, e)
       end
     end
 
```

How do you think that, @portante, @jcantrill, @richm ?

jcantrill:
Sample:
```
<label @ERROR>
  <match **> 
    @type file
    path /tmp/dlq
    utc 
    append true
  </match>
</label>
```

PR #399, opened 2018-04-16
Don't log full response on error

jcantrill:
This PR
* Removes logging of entire failed response for cases where debug is not enabled.
* Fluentd can get into a situation where it progressively degrades after failed response if it happens to also be scraping its own logs.
 
(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible


jcantrill:
cc @portante @richm

lib/fluent/plugin/elasticsearch_error_handler.rb
```diff
@@ -85,11 +85,17 @@ def handle_error(response)
     errors.each_key do |key|
       case key
       when 'out_of_memory_error'
-        raise ElasticsearchOutOfMemory, "Elasticsearch has exhausted its heap, retrying"
+        error_message = "Elasticsearch has exhausted its heap, retrying"
+        @plugin.log.error(error_message)
```
portante: This results in two log messages, can we remove this?

lib/fluent/plugin/elasticsearch_error_handler.rb
```diff
@@ -85,11 +85,17 @@ def handle_error(response)
...
+        raise ElasticsearchOutOfMemory, error_message
       when 'es_rejected_execution_exception'
-        raise BulkIndexQueueFull, "Bulk index queue is full, retrying"
+        error_message = "Bulk index queue is full, retrying"
+        @plugin.log.error(error_message)
```
portante: This results in two log messages, can we remove this?

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -452,7 +452,8 @@ def send_bulk(data)
       response = client.bulk body: data
       if response['errors']
         @error.handle_error(response)
-        log.error "Could not push log to Elasticsearch: #{response}"
+        log.error "Could not push log to Elasticsearch with Bulk Request.  Enable debug to see full response."
```
portante: Can we move these messages into the error handler?

lib/fluent/plugin/elasticsearch_error_handler.rb
```diff
@@ -85,11 +85,17 @@ def handle_error(response)
...
+        raise BulkIndexQueueFull, error_message
       else
-        raise ElasticsearchError, "Elasticsearch errors returned, retrying  #{response}"
+        error_message = "Elasticsearch errors returned, retrying  #{response}"
```
portante: Won't this format the string in all its glory even when we are not going to log it via `@plugin.log.debug()`?  Can we avoid that?

lib/fluent/plugin/out_elasticsearch.rb
cosmo0920: `Enable debug to see full response.` is not good for users.
Can we use `Use @log_level debug to see full response.` instead?

lib/fluent/plugin/elasticsearch_error_handler.rb
jcantrill: It will, but fluentd does not have a  'debug?' method so we either have to not log at all or take the pain.  Fluent will only log if its debug but we would get the string expansion

lib/fluent/plugin/out_elasticsearch.rb
jcantrill: @cosmo0920 I don't understand what you are suggesting here.  Aren't you saying to enable debug logging in the config which is what the message advises?

lib/fluent/plugin/elasticsearch_error_handler.rb
jcantrill: I stand corrected, they have an on_debug method that takes a block.

lib/fluent/plugin/out_elasticsearch.rb
cosmo0920: Yes. But it is optional in plugin code.
It should be in documentation.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -452,7 +452,8 @@ def send_bulk(data)
       response = client.bulk body: data
       if response['errors']
         @error.handle_error(response)
-        log.error "Could not push log to Elasticsearch: #{response}"
+        log.error "Could not push log to Elasticsearch with Bulk Request. Add '@log_level debug' to your config to see full response."
+        log.on_debug { log.debug "Failed bulk response: #{response}" }
```
richm: Do we need this here?  Looks like the handle_error method already logs the response in the error cases
I'm not sure we need the log.error here either

jcantrill:
@cosmo0920 Note I removed some of tests better targeted for the class under test

lib/fluent/plugin/elasticsearch_error_handler.rb
```diff
@@ -30,66 +25,52 @@ def handle_error(response)
...
       when 'out_of_memory_error'
-        raise ElasticsearchOutOfMemory, "Elasticsearch has exhausted its heap, retrying"
+        raise ElasticsearchOutOfMemory, 'Elasticsearch has exhausted its heap, retrying'
       when 'es_rejected_execution_exception'
-        raise BulkIndexQueueFull, "Bulk index queue is full, retrying"
+        raise BulkIndexQueueFull, 'Bulk index queue is full, retrying'
       else
-        raise ElasticsearchError, "Elasticsearch errors returned, retrying  #{response}"
+        @plugin.log.on_debug { @plugin.log.debug("Elasticsearch errors returned, retrying:  #{response}") }
+        raise ElasticsearchError, "Elasticsearch returned errors, retrying. Add '@log_level debug' to your config to see the full response"
```
cosmo0920: :+1:

PR #402, opened 2018-04-17
revert dlq to use router.emit_error_event instead

jcantrill:
This PR reverts #398 in favor of using router.emit_error_event which is more in line with what fluend suggests

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible


jcantrill:
cc @cosmo0920 @portante @richm

cosmo0920:
`test_write_message_with_bad_chunk` are failed.
We ought to resolve test failure before merging.

PR #437, opened 2018-06-20
Fix #434 bulk count

### LINKED ISSUE #434: Bulk errorhandling with mismatched response count leaves plugin stuck processing chunk
#### Problem
Bulk requests that fail have no mechanism to uniquely associate an entry in the request with an entry in the response.  Subsequently, we make the assumption the number of entries in the response are equal to those in the request and the returned order is the same as the order submitted.  The plugin can still become stuck retrying a chunk if the number of messages in the request do not equal the number in the response [1], ref [2].

The assumption here is that either Elasticsearch has an issue where it is not returning the proper number of responses or there is a bug in the plugin code.

I would propose one of the following:

* add a unique or ordinal id for the bulk submission to correlate to the response
or
* when there is a mismatch, send the whole chunk back through the retry

[1] https://github.com/uken/fluent-plugin-elasticsearch/blob/v0.12/lib/fluent/plugin/elasticsearch_error_handler.rb
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1562004#c30

#### Using Fluentd and ES plugin versions

* OS version  - rhel/centos7
* Fluentd v0.12
* ES plugin 1.16.1
* ES version 2.4.4

jcantrill:
This PR fixes #434 by:

* only increasing the bulk message count if a record is added to the bulk request
* adds configuration to emit an error event if record is missing _id; 
* adds debug logging when record is dropped because its missing _id

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible


jcantrill:
cc @richm @cosmo0920

cosmo0920:
Can using bundled `elasticsearch_genid` fiter plugin resolve missing _id field?

jcantrill:
@cosmo0920 it can and thats actually what we use in our recommended production configurations.  I believe the issue is this:
* fluent writes to file buffers
* fluent runs out of disk space and corrupts the record
* fluent successfully deserialize the message but its not complete (e.g. missing _id)
adding the ID is the last step we perform before trying to push to Elasticsearch.

As a side note, I'm unable to figure out why the plugin drops records that don't have an '_id' field for create operations.  Elasticsearch behavior, to my understanding, is it will generate an id if one is not provided.  This means the plugin, incorrectly, requires users to generate the id but the id generation is an optional configuration for the plugin.  IMO, it should either generate _id by default, or always submit records for create operations that do not have an id

cosmo0920:
> it can and thats actually what we use in our recommended production configurations.

Ah, I got it.

cosmo0920:
This PR is released as v1.17.0.

PR #445, opened 2018-07-20
Strictness scheme

cosmo0920:
We can validate scheme with `:enum` configuration type.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920:
We can also backport these changes into v0.12 branch.
Because Fluentd v0.12 supports `enum` type in config_param.

Issue #444, opened 2018-07-12
Support rest.action.multi.allow_explicit_index: false


dcrn:
When using the Elasticsearch Service on AWS, we have the ability to control user access depending on the index. This allows several services to utilize the same ES server without the possibility of one modifying data in the wrong index.

However, fluent-plugin-elasticsearch uses the `_bulk` endpoint without any index specified. Allowing one user to use the _bulk endpoint also gives them permissions to modify any index (as the index is specified in the request data), unless `rest.action.multi.allow_explicit_index` is set to `false`.

Setting the ElasticSearch configuration `rest.action.multi.allow_explicit_index` to `false` adds a requirement that the indices specified in the request data must match the index in the URL, e.g. `/{index}/{type}/_bulk`. After testing, it seems that it is currently not supported in fluent-plugin-elasticsearch. It would be great to have, especially for AWS users.

#### Using Fluentd and ES plugin versions

* OS version: Amazon Linux 2
* Fluentd v1.2.2
* ES plugin 2.10.3
* ES version: 6.2

cosmo0920:
I'm not familiar with AWS Elasticsearch Searvice.
Can ES plugin connect to this service?
If not so, this issue should be reported in https://github.com/atomita/fluent-plugin-aws-elasticsearch-service.

badyli:
We're using a simple proxy to add necessary headers, so you should treat it as normal ES instance (tested and no problems here).

Problem is, as described, the address of the endpoint. If we want to make our instance slighlty more secure we can use access policies to limit which index user can access. Currently the ES plugin is using main _bulk endpoint, e.g. ```http://localhost:9200/_bulk```. If the ```rest.action.multi.allow_explicit_index``` is set to ```false``` this action will fail. 
On the other hand, if the index operations in _bulk request always use the same index endpoint it should be possible to form the URL with index ```http://localhost:9200/_myindex/_bulk``` and process the operation without failing.

dcrn:
The proxy described above: https://github.com/abutaha/aws-es-proxy
This handles all of the authentication required for the AWS Elasticsearch Service, and essentially gives us a standard Elasticsearch endpoint.

cosmo0920:
fluent-plugin-elasticsearch does not use bulk URL explicitly.
So, you can send this feature request into [elasticsearch-ruby](https://github.com/elastic/elasticsearch-ruby).
Thanks.

dcrn:
[...]

PR #450, opened 2018-07-26
Add an option `include_index_in_url` to allow URL-based conrtrols

fujimotos:
This patch is the first cut at fixing the issue #444.

### What's this option all about?

This option tells Fluentd to manifest the name of the target index in
the request URL (rather than in the request body). For example:

    POST /logstash-20180501/_bulk

And without this option, Fluentd just uses the general endpoint:

    POST /_bulk

We need this option in order to enforce URL-based access controls on
Elasticsearch endpoints. For details, read the following article:

https://www.elastic.co/guide/en/elasticsearch/reference/1.5/url-access-control.html

### Checklist

- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/test/plugin/test_out_elasticsearch_dynamic.rb
+++ b/test/plugin/test_out_elasticsearch_dynamic.rb
@@ -990,4 +990,18 @@ class ElasticsearchOutputDynamic < Test::Unit::TestCase
     end
     assert(index_cmds[0].has_key?("create"))
   end
+
+  def test_include_index_in_url
+    stub_elastic_ping
+    stub_elastic('http://localhost:9200/logstash-2018.01.01/_bulk')
+
+    driver.configure("index_name logstash-2018.01.01
+                      include_index_in_url true")
+    driver.run(default_tag: 'test') do
+      driver.feed(sample_record)
+    end
+
+    assert_equal(index_cmds.length, 2)
+    assert_equal(index_cmds.first['index']['_index'], nil)
+  end
 end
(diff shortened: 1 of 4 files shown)


test/plugin/test_out_elasticsearch_dynamic.rb
```diff
@@ -990,4 +990,18 @@ def test_create_should_write_create_op
...
+    stub_elastic('http://localhost:9200/logstash-2018.01.01/_bulk')
+
+    driver.configure("index_name logstash-2018.01.01
+                      include_index_in_url true")
+    driver.run(default_tag: 'test') do
+      driver.feed(sample_record)
+    end
+
+    assert_equal(index_cmds.length, 2)
+    assert_equal(index_cmds.first['index']['_index'], nil)
```
cosmo0920: :+1:

fujimotos:
**NOTE**

I did not include the implementation for the normal elasticsearch plugin in this patch,
in order to keep this pull request simple to review.

Since the normal plugin has much more complicated logic to handle bulk requests,
we need a lot more work to integrate `include_index_in_url` to the plugin. So the
resulting patch would become barely reviewable if we include the diff.

Also I'm suspecting that the user who originally requested this feature was primarily
thinking of the dynamic plugin. So possibly, we can close the issue #444 just by merging
this patch.

cosmo0920:
Thanks for the patch, @fujimotos !
It's a perfect PR!! :heart_eyes:

cosmo0920:
I've released this change as v2.11.3.

PR #451, opened 2018-07-26
Implement the `include_index_in_url` option for out_elasticsearch

fujimotos:
This is a sequel to the previous patch #450.

### What's this patch?

With this patch applied, you can use this option with the normal elasticsearch plugin.

The nasty thing here is that we cannot assume all events in a single
chunk share the same index name even under a non-dynamic plugin. Why?
Because Elasticsearch's index names usually contain a timestamp (e.g.
'logstash-2018.01.01') and these timestamps are generated from each
record's time field.

This is why we had to convert `bulk_message` into a hash, and do all
the work to get the other methods working fine with it.

(Original issue: #444)

### Checklist

- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,8 +620,6 @@ content_type application/x-ndjson
 With this option set to true, Fluentd manifests the index name in the request URL (rather than in the request body).
 You can use this option to enforce an URL-based access control.
 
-This option is currently available only for `elasticsearch_dynamic`.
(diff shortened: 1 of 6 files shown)


cosmo0920:
I've released this change as v2.11.3.

PR #483, opened 2018-10-15
Avoid NoMethodError on unknown Elasticsearch error responses

Etsukata:
Current error handling can cause `NoMethodError` on unkown Elasticsearch
response because it assumes error responses include `type` or `reason`
members. But error responses of old Elasticsearch(1.X) can be just string.

Use `fetch('error', {})` so that it can avoid unrecoverable NoMethodError.

- [x] tests added
- [x] tests passing
- [-] README updated (if needed)
- [-] README Table of Contents updated (if needed)
- [x] History.md and `version` in gemspec are untouched
- [x] backward compatible
- [-] feature works in `elasticsearch_dynamic` (not required but recommended)

PR #487, opened 2018-10-16
Avoid NoMethodError on unknown Elasticsearch error responses

Etsukata:
Current error handling can cause `NoMethodError` on unkown Elasticsearch
response because it assumes error responses include `type` or `reason`
members. But error responses of old Elasticsearch(1.X) can be just string.

Use `fetch('error', {})` so that it can avoid unrecoverable NoMethodError.

- [x] tests added
- [x] tests passing
- [-] README updated (if needed)
- [-] README Table of Contents updated (if needed)
- [x] backward compatible
- [-] feature works in `elasticsearch_dynamic` (not required but recommended)

----

### Changes from https://github.com/uken/fluent-plugin-elasticsearch/pull/483

- Target branch changed / Rebased on `v0.12`


cosmo0920:
Thank you for your work! :+1:

cosmo0920:
I've released this change as v1.18.0.

Issue #485, opened 2018-10-15
Decoupling the custom template and rollover index creation


kumaravel29:
#### Problem
Feature Request - As per the feature request #355, we had added the capability to customize the templates. But the creation of rollover index(without any option to customize it) was introduced in it.
So whenever the customize template feature was used, default rollover was created in the format <logstash-default-{now/d}-000001> and a deflector alias **<template_name>-current** was assigned to the index.
I am trying to decouple this, so that users can have an option to customize this.
...

#### Expected Behavior or What you need to ask
I wanted this to be decoupled and give an option to the user for creation of index and customize it.
...

#### Using Fluentd and ES plugin versions

* OS version - Linux
* Fluentd v0.12 or v0.14/v1.0 
_gem 'fluentd' version '1.2.0'_
* ES plugin 2.x.y or 1.x.y
_{"time":"2018-10-15 13:02:00 +0530","level":"info","message":"parsing config file is succeeded path=\"/etc/td-agent/td-agent.conf\""}
{"time":"2018-10-15 13:02:01 +0530","level":"info","message":"starting fluentd-1.2.0 pid=591 ruby=\"2.1.10\""}
{"time":"2018-10-15 13:02:01 +0530","level":"info","message":"spawn command to main:  cmdline=[\"/opt/td-agent/embedded/bin/ruby\", \"-Eascii-8bit:ascii-8bit\", \"/usr/sbin/td-agent\", \"--log\", \"/var/log/td-agent/td-agent.log\", \"--daemon\", \"/var/run/td-agent/td-agent.pid\", \"--under-supervisor\"]"}
{"time":"2018-10-15 13:02:04 +0530","level":"info","message":"gem 'fluent-mixin-plaintextformatter' version '0.2.6'"}
{"time":"2018-10-15 13:02:04 +0530","level":"info","message":"gem 'fluent-plugin-copy_ex' version '0.1.0'"}
{"time":"2018-10-15 13:02:04 +0530","level":"info","message":"gem 'fluent-plugin-elasticsearch' version '2.11.11'"}
{"time":"2018-10-15 13:02:04 +0530","level":"info","message":"gem 'fluent-plugin-forest' version '0.3.3'"}
{"time":"2018-10-15 13:02:04 +0530","level":"info","message":"gem 'fluent-plugin-kafka' version '0.6.1'"}
[...]

Issue #355, opened 2018-02-05
Customize Elasticsearch template based on tag


kumaravel29:
In my setup, I create indices on the elasticsearch cluster based on the tag name. I have a requirement to set the alias name also for the indices according to the tag name
eg - I am having a tagname as ABC.input and my index is named as logstash-ABC-2018.01 and wanted to push a template for ABC and with alias ABC_alias

#### Using Fluentd and ES plugin versions

* OS version -  **CoreOS**
* Fluentd v0.12 or v0.14/v1.0 - **td-agent 0.14.21**
* ES version - **ES 2.3.4 and planning to move to 5.6**
* ES plugin
 **fluent-plugin-elasticsearch (1.10.0)
fluent-plugin-forest (0.3.3)**

cosmo0920:
Currently, ES plugin 1.x and 2.x install ES templates in `#configure`.
In `#configure`, we can install multiple ES templates with `templates` parameter's hash value.
Is this behavior is not expected for you?

Thanks.

kumaravel29:
The `templates` parameter is useful for multiple static templates. I have an use case for setting dynamic templates.
As in my earlier comment, I use the appId(`ABC`) for setting the tag name(`ABC.input.tag`) and then create index based on the appId (i.e 1st part of tag `ABC`) and hence the index name is logstash-abc-2018.01 with a weekly rotation. 
To try the same in my local setup, I updated the following line in elasticsearch_index_template.rb
`file_contents = IO.read(template_file).gsub(/\n/,'')` --> `file_contents = IO.read(template_file).gsub(/\n/,'').gsub(/##alias_name##/,name)` and my template had the string `##alias_name##`

We can add a new parameter something like custom_template which can have something like `{"string_1": "subs_value_1", "string_2": "subs_value_2"}` and then the same can be used to generate the template.

cosmo0920:
> We can add a new parameter something like custom_template which can have something like {"string_1": "subs_value_1", "string_2": "subs_value_2"} and then the same can be used to generate the template.

In your use case, should dynamic template parameters be fixed or determined in booting up time?
Or, they should be determined on each of bulk insert request?

kumaravel29:
I would go with dynamic template in booting up time because for each bulk insert request might have performance impact

cosmo0920:
Resolved on #431.

PR #486, opened 2018-10-15
Decoupling the custom template and rollover index creation #485

kumaravel29:
As per the feature request #355, we had added the capability to customize the templates. But the creation of rollover index(without any option to customize it) was introduced in it.
So whenever the customize template feature was used, default rollover was created in the format <logstash-default-{now/d}-000001> and a deflector alias <template_name>-current was assigned to the index.
I am trying to decouple this, so that users can have an option to customize this.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] README updated (if needed)
- [x] README Table of Contents updated (if needed)
- [x] History.md and `version` in gemspec are untouched
- [ ] backward compatible
- [ ] feature works in `elasticsearch_dynamic` (not required but recommended)

PR #488, opened 2018-10-16
Decoupling the custom template and rollover index creation #485

kumaravel29:
As per the feature request #355, we had added the capability to customize the templates. But the creation of rollover index(without any option to customize it) was introduced in it.
So whenever the customize template feature was used, default rollover was created in the format <logstash-default-{now/d}-000001> and a deflector alias <template_name>-current was assigned to the index.
I am trying to decouple this, so that users can have an option to customize this.

(check all that apply)
- [x] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,13 +341,14 @@ If `template_file` and `template_name` are set, then this parameter will be igno
 
 ### customize_template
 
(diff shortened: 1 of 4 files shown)


cosmo0920:
Closing this in favor of #486

Issue #506, opened 2018-11-26
Return statement ignores rollover setup when overwrite is true


jahmad-fareportal:
https://github.com/uken/fluent-plugin-elasticsearch/blob/573c39a757c2ef11449e327daa88f87e9961c4ee/lib/fluent/plugin/elasticsearch_index_template.rb#L76

cosmo0920:
Fill the issue template and write an issue description, please.

kumaravel29:
I will check on this and try fixing it.

cosmo0920:
Patch is merged. Closing.

PR #513, opened 2018-12-09
#506 Rollover index will be in effect in case of template overwrite also.

kumaravel29:
In case of template overwrite, the rollover settings were ignored. To handle this, the conditions have been modified. 
Reference - https://github.com/uken/fluent-plugin-elasticsearch/issues/506

(check all that apply)
- [x] tests added
- [ ] tests passing
- [ ] backward compatible


kumaravel29:
Thanks :)

cosmo0920:
I've released this change as 2.12.4.

PR #517, opened 2018-12-21
Disable connection retry mechanism and rely on Fluentd buffer flush retry mechanism

yann-soubeyrand:
Following a Slack discussion, I think that Elasticsearch pluging shouldn't try to implement a retry strategy but should instead rely on Fluentd buffer flush retry mechanism which is completely configurable: https://docs.fluentd.org/v1.0/articles/buffer-section#retries-parameters.

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -23,7 +23,6 @@ end
 module Fluent::Plugin
   class ElasticsearchOutput < Output
     class ConnectionFailure < StandardError; end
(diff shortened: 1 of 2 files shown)


yann-soubeyrand:
I'm working on fixing tests failures but I'm facing some difficulties. I'll keep you up to date ;-)

cosmo0920:
@yann-soubeyrand Could you take a look #516?
#516 stop to use ES plugin specific retrying mechanism and encourages to increase `flush_thread_count`.
I don'y want to change default hebavior where ``flush_thread_count` = 1.

PR #543, opened 2019-02-11
Don't attempt to connect to Elasticsearch in dry run mode

stevenjm:
This allows using `fluentd --dry-run` to test the fluentd config, without failing if there is no Elasticsearch server in the test environment.

(check all that apply)
- [ ] tests added
- [X] tests passing
- [X] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -173,20 +173,22 @@ EOC
 
       raise Fluent::ConfigError, "'max_retry_putting_template' must be positive number." if @max_retry_putting_template < 0
 
-      if @template_name && @template_file
-        retry_operate(@max_retry_putting_template) do
-          if @customize_template
-            if @rollover_index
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -173,7 +173,7 @@ def configure(conf)
 
       raise Fluent::ConfigError, "'max_retry_putting_template' must be positive number." if @max_retry_putting_template < 0
 
-      if @template_name && @template_file
+      if @template_name && @template_file && !Fluent::Engine.dry_run_mode
```
cosmo0920: Should we also add `&& !Fluent::Engine.dry_run_mode` clause for the following elsif statement?

```diff
+      elsif @templates && !Fluent::Engine.dry_run_mode
-      elsif @templates
         retry_operate(@max_retry_putting_template) do
           templates_hash_install(@templates, @template_overwrite)
         end
       end
```

stevenjm: Indeed, I've pushed a new version that just wraps that whole if/elif block in `if !Fluent::Engine.dry_run_mode`.

PR #557, opened 2019-03-07
add new option to suppress doc wrapping 

photom:
### What I want to do
To post script record as described at elasticsearch official docs.
Please see below.
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update

### Problem
fluent-plugin-elasticsearch wraps update record with 'doc' and can not handle script.
https://github.com/uken/fluent-plugin-elasticsearch/blob/master/lib/fluent/plugin/out_elasticsearch.rb#L471

### What I did
Add option that suppresses doc wrapping and allows record to pass through as is.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Current maintainers: @cosmo0920
   + [Placeholders](#placeholders)
   + [Multi workers](#multi-workers)
   + [log_es_400_reason](#log_es_400_reason)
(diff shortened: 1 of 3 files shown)


photom:
Thanks!

PR #562, opened 2019-03-19
Add ignore_exceptions config

ialidzhikov:
As discussed in [fluent/fluentd#2280](https://github.com/fluent/fluentd/issues/2280) this PR adds possibility to ignore configurable exceptions. This is useful in cases when you have dynamic host config like:
```
<match **>
  @type elasticsearch
  host ${key1}
 
  // omitted
</match>
```

In this config when you send dynamically to n elastic instances, and if some of them are unavailable this will result in:
```
failed to flush the buffer, and hit limit for retries. dropping all chunks in the buffer queue.
```
and in this buffer queue there could be some potentially successful chunks. 
The mechanism now check the exception that occurred in `ignore_exceptions` list and if a match is found, then the chunks is discarded and the retry mechanism is not called.

DESCRIPTION HERE

(check all that apply)
- [x] tests added
- [ ] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ Current maintainers: @cosmo0920
   + [Multi workers](#multi-workers)
   + [log_es_400_reason](#log_es_400_reason)
   + [suppress_doc_wrap](#suppress_doc_wrap)
+  + [ignore_exceptions](#ignore_exceptions)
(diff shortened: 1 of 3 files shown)


ialidzhikov:
Thank you!

Issue #423, opened 2018-05-21
no implicit conversion of Fluent::EventTime into String


mattheworiordan:
#### Problem

I have a huge volume of time parsing errors since upgrading this plugin, and am at a bit of a loss as to how I can get to the bottom of the issue as I don't really understand the tool chain for debugging fluentd and plugins.

The error I am seeing at a rate of literally 50+ per second per `fluentd` process (I run 8):

```
2018-05-21 21:52:33 +0000 [warn]: #0 dump an error event: error_class=TypeError error="no implicit conversion of Fluent::EventTime into String" location="/var/lib/gems/2.3.0/gems/fluent-plugin-elasticsearch-2.10.1/lib/fluent/plugin/out_elasticsearch.rb:222:in `parse'" tag="Fluent::ElasticsearchOutput::TimeParser.error" time=#<Fluent::EventTime:0x00007f41bab45ed0 @sec=1526939553, @nsec=948145538> record={"tag"=>"fluent.warn", "time"=>#<Fluent::EventTime:0x00007f41bab45048 @sec=1526939431, @nsec=117101150>, "format"=>nil, "value"=>#<Fluent::EventTime:0x00007f41bab44648 @sec=1526939431, @nsec=117075662>}
```

#### Steps to replicate

The problem I have is that I don\t know how to work out which of my log sources is causing this problem.  Any advice on how to enabled debugging mode, or perhaps output the log lines that are causing these failures, will help me very quickly figure out what is causing the issue.

The `fluentd.conf` is at https://gist.github.com/mattheworiordan/08c16983b62a9a84f01e420001cf50db.  Note it's an ERB file, so all `<%=%>` get replaced at runtime.

#### Expected Behavior or What you need to ask

I would like to know which log entries are causing this error (lots of logs are coming through) so that I can help fix it.

#### Using Fluentd and ES plugin versions

* OS version - Ubuntu 14 in a Docker container
* Fluentd v0.14.25
* ES plugin 2.10.1
```
$ bundle list
Gems included by the bundle:
  * bundler (1.16.0)
  * cool.io (1.5.3)
  * elasticsearch (6.0.2)
  * elasticsearch-api (6.0.2)
  * elasticsearch-transport (6.0.2)
  * excon (0.62.0)
  * faraday (0.15.1)
  * fluent-plugin-color-stripper (0.0.3 a1ac8dd)
  * fluent-plugin-elasticsearch (1.16.1)
[...]

PR #570, opened 2019-04-02
Handle non-String value on parse_time

cosmo0920:
Related to #423.

Could you take a look, @mattheworiordan @pathikritmodak @bderickson  ?

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -368,10 +368,7 @@ EOC
         end
       else
         Proc.new { |value|
-          if value.is_a?(Numeric)
-            numeric_time_parser = Fluent::NumericTimeParser.new(:float)
-            value = Time.at(numeric_time_parser.parse(value).to_r).strftime("%Y-%m-%d %H:%M:%S.%N %z")
-          end
+          value = convert_numeric_time_into_string(value) if value.is_a?(Numeric)
           DateTime.parse(value)
         }
       end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -366,7 +367,13 @@ def create_time_parser
           Proc.new { |value| DateTime.strptime(value, @time_key_format) }
```
bderickson: I think you're gonna need similar handling here as well. While I was hunting down this problem, I tried providing a `time_key_format` of `%s.%N`, but that also failed for the same reason.

bderickson: I didn't use the `strptime` gem, but I imagine it'll need similar checking (line 361).

cosmo0920: I wish you showed me your Fluentd configuration and reproducible steps.
Just commenting not work for me is totally useless to solve issue, thanks!!

bderickson: The only relevant variables here are `@time_key_format` and the value that gets passed into the `Proc` object, so I thought the `time_key_format` configuration option that I provided was enough. The rest of any config file I could provide wouldn't matter.

Since there are three code paths in this function and you're only covering one, I think it would make more sense to pull lines 372 and 373 into a separate function, maybe named `convert_numeric_time_to_string`. Then update the three `Proc` blocks in the `create_time_parser` function to:
```
Proc.new { |value|
  if value.is_a?(Numeric)
    value = convert_numeric_time_to_string(value)
  end
  ...
```

cosmo0920: Yeah, it is reasonable for me.

cosmo0920: I've committed to follow your suggestion with a few improving.
Could you check again?

PR #573, opened 2019-04-14
Comparing DEFAULT_TYPE_NAME_ES_7x to target_type instead of type_name

jjanier:
This produce a large amount of unexpected warning logs.

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -667,7 +667,7 @@ EOC
           target_type = '_doc'.freeze
         end
       else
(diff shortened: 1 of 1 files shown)


chenjienn:
@cosmo0920  i met this problem too , but how could i update plugin in td-agent?

cosmo0920:
Try `td-agent-gem update fluent-plugin-elasticsearch -v 3.5.0`.

But you should also update elasticsearch ruby client with:

`td-agent-gem update elasticsearch -v 7.0.0`.

ref: https://docs.fluentd.org/v1.0/articles/plugin-management#if-using-td-agent,-use-/usr/sbin/td-agent-gem

chenjienn:
@cosmo0920 thanks !!!

PR #579, opened 2019-04-18
Add fail_on_putting_template_retry_exceed config

ialidzhikov:
Currently the elastic plugin with the following config
```
<match elastic.**>
  @type elasticsearch
  [...]
  template_name test_name
  template_file /test_template
  [...]
</match>
```
tries to apply the index templates in the `configure` step. If the apply step fails, the plugin starts retrying before to fail after `max_retry_putting_template` and fluentd fails to start.

You could also have another output plugins such as:
```
<match foo.**>
  @type stdout
  @id stdout_foo
</match>

<match bar.**>
  @type stdout
  @id stdout_bar
</match>
```
and probably you will want to have fluentd running with the other output plugins even without the elastic templates.

The PR introduces `fail_on_putting_template_retry_exceed` config that can ignore  `max_retry_putting_template` exceed.

DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Current maintainers: @cosmo0920
   + [index_prefix](#index_prefix)
   + [templates](#templates)
   + [max_retry_putting_template](#max_retry_putting_template)
+  + [fail_on_putting_template_retry_exceed](#fail_on_putting_template_retry_exceed)
   + [max_retry_get_es_version](#max_retry_get_es_version)
   + [request_timeout](#request_timeout)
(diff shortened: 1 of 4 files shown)


ialidzhikov:
Thank you!

cosmo0920:
@ialidzhikov I've released this change as v3.4.3.
Sorry to being late to release. Thanks for your contributions. :smile:

Issue #622, opened 2019-08-09
Timezone-dependent unit test


JBury-rms:
(check apply)
- [X] read [the contribution guideline](https://github.com/uken/fluent-plugin-elasticsearch/blob/master/CONTRIBUTING.md)

#### Problem
https://github.com/uken/fluent-plugin-elasticsearch/blob/master/test/plugin/test_out_elasticsearch.rb#L1792

This test sometimes fails if run in a timezone that is effected by daylight saving's time.  In my case, the timezone of DateTime.now is PDT (-07:00), but the hard-coded date-time would have the PST timezone (-08:00).  Since we're manually changing timezone offsets based on the current timestamp, the test will fail depending on whether we're in DST or not.

#### Steps to replicate
1) Create a fresh copy of this repo
2) Set your timezone to PDT
3) Run the unit tests

#### Expected Behavior or What you need to ask
All tests should pass

#### Using Fluentd and ES plugin versions
Shouldn't really matter - this is a timezone issue, not a versioning issue.

PR #623, opened 2019-08-09
Fix DST-breaking unit test

JBury-rms:
This is a fix for https://github.com/uken/fluent-plugin-elasticsearch/issues/622

Currently the /test/plugin/test_out_elasticsearch.rb testcase
test_uses_custom_time_key_with_float_record_and_format breaks
depending on whether or not the timezone generated by Time.now
is the same as the timezone that would be used on 2001-02-03.

This is a simple fix that bases the current_zone_offset off of
2001-02-03 instead of Time.now.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920:
Oh, I didn't notice about it.
Good catch. Thank you!!

PR #627, opened 2019-08-21
Fix error message for `max_retry_putting_template`

kenrota:
Fixed error message for `max_retry_putting_template`.
Zero is not positive number.

```ruby
0.positive? # => false
0.negative? # => false
```

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible


kenrota:
Thank you!

PR #635, opened 2019-08-24
Fix arguments order of `assert_equal`

kenrota:
Fix arguments order of `assert_equal`.
First argument is expected result.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible


kenrota:
Thanks!

PR #636, opened 2019-08-27
Validate `user` and `password` early

kenrota:
- Check `nil` more explicitly by using `.nil?`
- Fix typo. 'pasword' => 'password'
- Validate `user` and `password` early, before connection.

(check all that apply)
- [x] tests added
- [x] tests passing
- [ ] backward compatible


kenrota:
Thanks!

Issue #565, opened 2019-03-21
Index template does not work with dynamic host


ialidzhikov:
#### Problem
Index template does not work with dynamic host.

#### Steps to replicate

1. Start fluentd with conf:

```
<source>
  @type forward
  @id forward_input
</source>

<match elastic.**>
  @id elasticsearch_dynamic
  @type elasticsearch

  host ${key1}

  port 9200
  logstash_format true

  reconnect_on_error true

  template_name gardener
  template_file /Users/i331370/default.template

  <buffer tag, key1, time>
    @type memory
  </buffer>
</match>
```

2. Ensure that the start fails with:

```
2019-03-21 15:34:12 +0200 [error]: #0 unexpected error error_class=URI::InvalidURIError error="bad URI(is not URI?): http://${key1}:9200"

// stacktrace omitted
```

#### Expected Behavior or What you need to ask
Index template to work with dynamically configured host and to create the index template for each host.

#### Using Fluentd and ES plugin versions

* OS version
* Bare Metal or within Docker or Kubernetes or others?
* Fluentd v0.12 or v0.14/v1.0 - `fluentd 1.4.0`
  * paste result of ``fluentd --version`` or ``td-agent --version``
* ES plugin 3.x.y/2.x.y or 1.x.y - `3.3.3`
  * paste boot log of fluentd or td-agent
  * paste result of ``fluent-gem list``, ``td-agent-gem list`` or your Gemfile.lock
* ES version (optional)
* ES template(s) (optional)

cosmo0920:
This is because index template installation is working on #configure phase not #write phase.
This is plugin limitation and we cannot handle #configure phase because dynamic host feature is non-determistic on #configure.

Do you have any incredible solution?
I have no idea for now. :cry:

cosmo0920:
This is known issue for elasticsearch_dynamic plugin.
out_elasticsearch inherits ES dynamic plugin behavior.
If we want to install ES templates, we should specify ES host candidates in #configura phase.

cosmo0920:
I decided that this issue should be treated as "plugin limitation".
Fluentd philosophy requests that plugin's `#configure` phase handles deterministic configurations and validates its configurations.

But to support your feature requests, it needs to supply host candidates information to extract placeholders with created metadata where `#configure`.

If this feature will be supported, `#configure` phase too long to wait Fluentd boot, I think.
(In your usecase seems to have to handle 100+ servers.)

[...]

PR #651, opened 2019-10-22
Support index lifecycle management

fpytloun:
DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -197,28 +197,17 @@ EOC
         raise Fluent::ConfigError, "host placeholder and template installation are exclusive features."
       end
 
-      if !Fluent::Engine.dry_run_mode
-        if @template_name && @template_file
-          retry_operate(@max_retry_putting_template, @fail_on_putting_template_retry_exceed) do
-            if @rollover_index
-              raise Fluent::ConfigError, "'deflector_alias' must be provided if 'rollover_index' is set true ." if not @deflector_alias
-            end
-            if @enable_ilm
-              raise Fluent::ConfigError, "'rollover_index' and 'deflector_alias' must be provided if 'enable_ilm' is set true ." if !@deflector_alias &&!@deflector_alias
-            end
-
-            verify_ilm_working if @enable_ilm
-            if @customize_template
-              template_custom_install(@template_name, @template_file, @template_overwrite, @customize_template, @enable_ilm, @deflector_alias, @ilm_policy_id)
-            else
-              template_install(@template_name, @template_file, @template_overwrite, @enable_ilm, @deflector_alias, @ilm_policy_id)
-            end
-            create_rollover_alias(@index_prefix, @rollover_index, @deflector_alias, @application_name, @index_date_pattern, @enable_ilm, @ilm_policy_id, @ilm_policy)
-          end
-        elsif @templates
-          retry_operate(@max_retry_putting_template, @fail_on_putting_template_retry_exceed) do
-            templates_hash_install(@templates, @template_overwrite)
-          end
+      if @template_name && @template_file
+        if @rollover_index
+          raise Fluent::ConfigError, "'deflector_alias' must be provided if 'rollover_index' is set true ." if not @deflector_alias
+        end
+        if @enable_ilm
+          raise Fluent::ConfigError, "'rollover_index' and 'deflector_alias' must be provided if 'enable_ilm' is set true ." if !@deflector_alias &&!@deflector_alias
+        end
+        verify_ilm_working if @enable_ilm
+      elsif @templates
+        retry_operate(@max_retry_putting_template, @fail_on_putting_template_retry_exceed) do
+          templates_hash_install(@templates, @template_overwrite)
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -197,28 +198,17 @@ def configure(conf)
...
-            else
-              template_install(@template_name, @template_file, @template_overwrite, @enable_ilm, @deflector_alias, @ilm_policy_id)
-            end
-            create_rollover_alias(@index_prefix, @rollover_index, @deflector_alias, @application_name, @index_date_pattern, @enable_ilm, @ilm_policy_id, @ilm_policy)
-          end
-        elsif @templates
-          retry_operate(@max_retry_putting_template, @fail_on_putting_template_retry_exceed) do
-            templates_hash_install(@templates, @template_overwrite)
-          end
+      if @template_name && @template_file
```
cosmo0920: Looks reasonable.
But, could you use `!Fluent::Engine.dry_run_mode` for using network code?
Fluentd `#configure` is run twice as dry run mode and normal mode.
I don't want to change previous behavior.

fpytloun: Will fix, I forgot this conditional when moving the code.

fpytloun:
@cosmo0920 I am thinking about one thing - every bulk write will now check presence of template and index alias. I don't see any performance impact so far on my environment but anyway I am thinking about some way how to avoid this and do this operation only once for each alias. Is it possible to have some caching or something? Or do you have better idea?

cosmo0920:
> Is it possible to have some caching or something? Or do you have better idea?

I think that elasticsearch_dynamic's `dynamic_config` like caching mechanism can handle this issue. But, too old caching template clause should be dropped due to memory consumption.
If this PR doesn't makes heavily ES cluster operation, we can delay to tackle it on later.

lib/fluent/plugin/out_elasticsearch.rb
cosmo0920: I've fixed test failures and this comment fix.
Could you confirm them?

fpytloun:
> > Is it possible to have some caching or something? Or do you have better idea?
> 
> I think that elasticsearch_dynamic's `dynamic_config` like caching mechanism can handle this issue. But, too old caching template clause should be dropped due to memory consumption.
> If this PR doesn't make heavily ES cluster operation, we can delay to tackle it on later.

Ok, added caching of index aliases, tested and seems it's working in my env. Don't suppose it should hold a lot of data to cause some memory leak issue as only single entry is added for each index alias.

cosmo0920:
> > > Is it possible to have some caching or something? Or do you have better idea?
> > 
> > 
> > I think that elasticsearch_dynamic's `dynamic_config` like caching mechanism can handle this issue. But, too old caching template clause should be dropped due to memory consumption.
> > If this PR doesn't make heavily ES cluster operation, we can delay to tackle it on later.
> 
> Ok, added caching of index aliases, tested and seems it's working in my env. Don't suppose it should hold a lot of data to cause some memory leak issue as only single entry is added for each index alias.

Cool. I've added two testcases for your change.
And this PR also will fix the https://github.com/uken/fluent-plugin-elasticsearch/issues/565 with minor change. I'll work for it after merging this PR.

fpytloun:
Thank you! :-)

Just in case someone wants use this feature before it's merge into elasticsearch-ruby, here's example Dockerfile that I am using:

```
FROM fluent/fluentd:v1.7.3-1.0

USER root
RUN apk add --update --virtual .build-deps \
        sudo build-base ruby-dev ruby-rake ruby-bundler zlib-dev curl-dev git \
 && gem install \
        fluent-plugin-azure-storage-append-blob \
        fluent-plugin-concat \
        fluent-plugin-detect-exceptions \
        #fluent-plugin-elasticsearch \
        fluent-plugin-flatten-hash \
        fluent-plugin-multi-format-parser \
        fluent-plugin-prometheus \
        fluent-plugin-rename-key \
        fluent-plugin-rewrite-tag-filter \
        fluent-plugin-secure-forward \
        fluent-plugin-throttle \
        aws-sdk-sqs \
        yajl-ruby \
 # XXX: Install patched elasticsearch-ruby
 && git clone -b support-ilm https://github.com/cosmo0920/elasticsearch-ruby.git /usr/src/elasticsearch-ruby \
 && cd /usr/src/elasticsearch-ruby && bundle install && find . -name "*.gemspec"|cut -d / -f 2 | while read i; do cd $i && cp LICENSE.txt LICENSE && rake install; cd ..; done \
 # XXX: Install patched elasticsearch plugin to support ILM
 && git clone -b support-index-lifecycle-management https://github.com/uken/fluent-plugin-elasticsearch.git /usr/src/fluent-plugin-elasticsearch \
 && cd /usr/src/fluent-plugin-elasticsearch && rake install \
 && cd / && rm -rf /usr/src/* \
 && gem sources --clear-all \
 && apk del .build-deps \
 && rm -rf /var/cache/apk/* \
           /home/fluent/.gem/ruby/2.3.0/cache/*.gem

ADD plugins /fluentd/plugins
#USER fluent

EXPOSE 24224

COPY entrypoint.sh /bin/
ENTRYPOINT ["tini",  "--", "/bin/entrypoint.sh"]
CMD ["fluentd"]
```

PR #659, opened 2019-11-02
Merge Support index lifecycle management into master

cosmo0920:
This PR is merge PR to support index lifecycle management.
elasticsearch-ruby v7.4.0 will be released in the next week.
So, we can release index lifecycle management supported Elasticsearch plugin after that.

Fixes #565 
Fixes #540 

(check all that apply)
- [x] tests added
- [x] tests passing
- [ ] README updated (if needed)
- [ ] README Table of Contents updated (if needed)
- [x] History.md and `version` in gemspec are untouched
- [x] backward compatible
- [ ] feature works in `elasticsearch_dynamic` (not required but recommended)

PR #663, opened 2019-11-06
Issue655

filippovdenis:
Compression feature is added. You can set compression_level option choosing one of the values: 'default_compression', 'best_compression', 'best_speed'. There is also 'no_compression' value when compression isn't needed which is used by default.

- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -14,6 +14,7 @@ require 'fluent/event'
 require 'fluent/error'
 require 'fluent/time'
 require 'fluent/log-ext'
+require 'zlib'
 require_relative 'elasticsearch_constants'
 require_relative 'elasticsearch_error'
 require_relative 'elasticsearch_error_handler'
@@ -139,6 +140,7 @@ EOC
     config_param :ignore_exceptions, :array, :default => [], value_type: :string, :desc => "Ignorable exception list"
     config_param :exception_backup, :bool, :default => true, :desc => "Chunk backup flag when ignore exception occured"
     config_param :bulk_message_request_threshold, :size, :default => TARGET_BULK_BYTES
+    config_param :compression_level, :enum, {list: [:no_compression, :best_speed, :best_compression, :default_compression], :default => :no_compression}
 
     config_section :buffer do
       config_set_default :@type, DEFAULT_BUFFER_TYPE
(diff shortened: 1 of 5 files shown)


cosmo0920:
Thanks for scintillating your work!
But, current patch is conflicted due to big patch (#659).

Could you rebase of master?

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -309,6 +311,18 @@ class << self
           alias_method :split_request?, :split_request_size_check?
         end
       end
+
+      version_arr = Elasticsearch::Transport::VERSION.split('.')
+
+      if (version_arr[0].to_i < 7) || (version_arr[0].to_i == 7 && version_arr[1].to_i < 2)
+        if compression
+          raise Fluent::ConfigError, <<-EOC
+            Cannot use compression with elasticsearch-transport plugin version < 7.2.0
```
cosmo0920: It is good for new users. Nice. :+1:

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -420,7 +451,10 @@ def client(host = nil)
         if local_reload_connections && @reload_after > DEFAULT_RELOAD_AFTER
           local_reload_connections = @reload_after
         end
-        headers = { 'Content-Type' => @content_type.to_s }.merge(@custom_headers)
+
+        gzip_headers = if compression then {'Content-Encoding' => 'gzip'} else {} end
```
cosmo0920: Could you break single `if .. else .. end` line into multiple lines?

lib/fluent/plugin/out_elasticsearch_dynamic.rb
```diff
@@ -44,21 +44,24 @@ def client(host = nil)
       @_es ||= begin
         @current_config = connection_options[:hosts].clone
         adapter_conf = lambda {|f| f.adapter @http_backend, @backend_options }
+        gzip_headers = if compression then {'Content-Encoding' => 'gzip'} else {} end
```
cosmo0920: ditto.

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -309,6 +311,18 @@ class << self
           alias_method :split_request?, :split_request_size_check?
         end
       end
+
+      version_arr = Elasticsearch::Transport::VERSION.split('.')
+
+      if (version_arr[0].to_i < 7) || (version_arr[0].to_i == 7 && version_arr[1].to_i < 2)
```
cosmo0920: Can we use `Elasticsearch::Transport::VERSION < "7.2.0"` here?

filippovdenis: Nope
When Transport version will be 10.0.0 it won't work, we'll get an error

cosmo0920: Oh, I've missed the pitfall. Thanks. :smile:

PR #664, opened 2019-11-06
Support compression feature

filippovdenis:
Compression feature is added. You can set compression_level option choosing one of the values: 'default_compression', 'best_compression', 'best_speed'. There is also 'no_compression' value when compression isn't needed which is used by default.

- [x] tests added
- [ ] tests passing
- [x] backward compatible

### DIFF
--- a/test/plugin/test_out_elasticsearch.rb
+++ b/test/plugin/test_out_elasticsearch.rb
@@ -1831,7 +1831,7 @@ class ElasticsearchOutput < Test::Unit::TestCase
     compressed_body = gzip(bodystr, Zlib::DEFAULT_COMPRESSION)
 
     elastic_request = stub_request(:post, "http://localhost:9200/_bulk").
-        to_return(:status => 200, :headers => {"Content-Type": "Application/json"}, :body => compressed_body)
+        to_return(:status => 200, :headers => {'Content-Type' => 'Application/json'}, :body => compressed_body)
 
     driver(config)
     driver.run(default_tag: 'test') do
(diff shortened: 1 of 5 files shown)


test/plugin/test_out_elasticsearch.rb
```diff
@@ -1762,6 +1800,47 @@ def test_writes_to_default_index(data)
...
+                "_type"  : "fluentd"
+              }
+            }
+           ]
+        })
+
+    compressed_body = gzip(bodystr, Zlib::DEFAULT_COMPRESSION)
+
+    elastic_request = stub_request(:post, "http://localhost:9200/_bulk").
+        to_return(:status => 200, :headers => {"Content-Type": "Application/json"}, :body => compressed_body)
```
cosmo0920: Could you use rocket hash style?
ES plugin still support Ruby 2.1.
JSON like Hash style had been supported since Ruby 2.2

test/plugin/test_out_elasticsearch_dynamic.rb
```diff
@@ -305,6 +343,46 @@ def test_writes_to_default_index
...
+                "_type"  : "fluentd"
+              }
+            }
+           ]
+        })
+
+    compressed_body = gzip(bodystr, Zlib::DEFAULT_COMPRESSION)
+
+    elastic_request = stub_request(:post, "http://localhost:9200/_bulk").
+        to_return(:status => 200, :headers => {"Content-Type": "Application/json"}, :body => compressed_body)
```
cosmo0920: ditto.

test/plugin/test_out_elasticsearch.rb
filippovdenis: fixed

test/plugin/test_out_elasticsearch_dynamic.rb
filippovdenis: fixed

PR #684, opened 2019-12-13
Retry upsert on recoverable error. (porting to v0.12) #667

photom:
This patch is #667 and #583 porting to v0.12.

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920:
I'd published this change as v1.18.2.
Thank you for your work, @photom !

PR #692, opened 2020-01-11
Reduce log noise when not using rollover_index

JamesJJ:
Humbly suggest that this message should be at "debug" level instead of info.

The message is highlighting that we did not do something, because the configuration is not set to do that.  For rollover_index case, I think this matches expectations, so we should not need not emphasize so much.

The function is called often, here: https://github.com/uken/fluent-plugin-elasticsearch/blob/7282f515e4be2d18709ac6b516de67acc784ff3d/lib/fluent/plugin/out_elasticsearch.rb#L842


cosmo0920:
CI failure is not related this change.
I’ll fix them later.

PR #718, opened 2020-02-17
add logstash_dateformat as placeholder

tantra35:
Sometimes its veru useful to interpret logstash_dateformat as placeholder

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -656,6 +656,7 @@ EOC
 
     def expand_placeholders(chunk)
       logstash_prefix = extract_placeholders(@logstash_prefix, chunk)
+      logstash_dateformat = extract_placeholders(@logstash_dateformat, chunk)
       index_name = extract_placeholders(@index_name, chunk)
       if @type_name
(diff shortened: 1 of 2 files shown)


tantra35:
@cosmo0920 we have multiple tags, and we saving them to different indexes depend on  `tag` name, now we want also separate this indexes with different rollover policy, so one indexes will rotated every day, some indexes will rotate every month

so we have follow base filter

```
<filter **>
  @type record_transformer

  <record>
    indexformat %Y.%m.%d
  </record>
</filter>
```

and then for some tags we redefine `indexformat` like this:

```
<filter app.**>
  @type record_transformer

  <record>
    indexname automation.${tag}.atf01
    indexformat %Y.%m
  </record>
</filter>
 ```

and then we have only one `map` section which allow to us cover all possibilities in index creation. Before this patch we need multiple `map` section to satisfy multiple  values in `logstash_dateformat`, this is very inconvenient

tantra35:
@cosmo0920 We not so familiar with ruby, but we will try

cosmo0920:
@tantra35 If you won't fix for requested changes, I'll try to fix them and implement tests in the next week.

Issue #732, opened 2020-03-16
Fluentd on K8s stopped flushing logs to Elastic


nanorobocop:
(check apply)
- [x] read [the contribution guideline](https://github.com/uken/fluent-plugin-elasticsearch/blob/master/CONTRIBUTING.md)

#### Problem

We have Fluentd running in Daemonset (using fluentd-kubernetes-daemonset).
At some point almost all instances of Fluentd stop flushing their queue.

<img width="822" alt="screenshot" src="https://user-images.githubusercontent.com/3330092/76740038-8ca8d000-67b0-11ea-9bb7-e45d836e246c.png">

I checked related issues like https://github.com/uken/fluent-plugin-elasticsearch/issues/525 and https://github.com/rancher/rancher/issues/25819 and https://github.com/uken/fluent-plugin-elasticsearch/issues/600, but no luck.

Elastic is behind load balancer, so we set following setting:
```
      reload_connections false
      reconnect_on_error true
      reload_on_failure false
```

No errors or warning messages appears in fluentd logs.

I checked lsof: buffer and queue files are opened by ruby process:
```
root@fluentd-5m67p:/home/fluent# ps auxf
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root       6811  0.0  0.0   9744  2528 ?        Ss   08:53   0:00 bash
root       6818  0.0  0.0  13244  1784 ?        R+   08:53   0:00  \_ ps auxf
root          1  0.0  0.0   8288   168 ?        Ss   Mar12   0:07 tini -- /fluentd/entrypoint.sh
root          7  0.0  0.0 147120 59736 ?        Sl   Mar12   1:34 ruby /fluentd/vendor/bundle/ruby/2.6.0/bin/fluentd -c /fluentd/etc/fluentd.conf -p /fluentd/plugins --gemfile /fluentd/Gemfile -r /fluentd
root         19  0.2  0.1 391980 92724 ?        Sl   Mar12  16:08  \_ /usr/local/bin/ruby -Eascii-8bit:ascii-8bit /fluentd/vendor/bundle/ruby/2.6.0/bin/fluentd -c /fluentd/etc/fluentd.conf -p /fluentd/plu
root@fluentd-5m67p:/home/fluent# lsof | grep /var/log/fluentd | head
[...]

PR #733, opened 2020-03-22
Added http_backend_excon_nonblock config in out_elasticsearch

ukinau:
**Description of http_backend_excon_nonblock:**
 - If http_backend_excon_nonblock is true, Farday create Excon instance
 with nonblocking option which perform http request in nonblocking
 fassion
 - If http_backend_excon_nonblock is false, Fardday create Excon
 instance with blocking option which perform http request in blocking
 fassion

**Special Note about http_backend_excon_nonblock:**
If user use jRuby, you need to consider to configure http_backend_excon_nonblock=false.
Since you may experince some issue.
- https://github.com/geemus/excon/issues/106
- https://github.com/jruby/jruby-ossl/issues/19

But if user is not using jRuby, you'd better to configure
http_backend_excon_nonblock=true to prvent fluent-plugine-elasticsearch
from stucking

**Background:**
The older version of v1.0.0(latest as of 23 March 2020)
in Fardday(https://github.com/lostisland/farada),
When we use excon(https://github.com/excon/excon) as a adapter for
https, Faraday force to configure Excon to disable nonblock due to
following jRuby specific problem
- https://github.com/geemus/excon/issues/106
- https://github.com/jruby/jruby-ossl/issues/19

That's why currently when fluent-plugin-elasticsearch plugin connect to
Elastic Search, If it use http, Farday generate Excon instance with
nonblocking=true(default of Excon), but in the case of https, Farady
generate Excon instance with nonblocking=falase

This actually cause the problem making Farday HTTP Client stuck when SSL
Server(Elastic Search in current context) stuck during SSL Handshake
since Excon SSLSocket can not respecting to timeout, and rely on timeout
of SSL::SSLSocket in Ruby Standard Library which have exaclty process-hang
problem.

That's why sometimes flusth_thread of fluent-plugin-elasticsearch got
stuck when Elastic Search had something problem during SSL Session
handling.

Root cause of this problem is actually implementation of SSL::SSLSocket
in Ruby Standard Library(https://redmine.ruby-lang.org/issues/15729),
so better to fix Ruby side but it will take a time, and require ruby
version up, so better to provide the way to perform workaround.

(check all that apply)
- [ ] tests added
- [X] tests passing
- [X] README updated (if needed)
- [X] README Table of Contents updated (if needed)
- [X] History.md and `version` in gemspec are untouched
- [X] backward compatible
- [X] feature works in `elasticsearch_dynamic` (not required but recommended)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Current maintainers: @cosmo0920
   + [content_type](#content_type)
   + [include_index_in_url](#include_index_in_url)
   + [http_backend](#http_backend)
+  + [http_backend_excon_nonblock](#http_backend_excon_nonblock)
   + [prefer_oj_serializer](#prefer_oj_serializer)
   + [compression_level](#compression_level)
   + [Client/host certificate options](#clienthost-certificate-options)
@@ -814,6 +815,21 @@ Default value is `excon` which is default http_backend of elasticsearch plugin.
 
 ```
 http_backend typhoeus
+```
+
+### http_backend_excon_nonblock
+
+With `http_backend_excon_nonblock false`, elasticsearch plugin use excon with nonblock=false.
+If you use elasticsearch plugin with jRuby for https, you may need to consider to set `false` to avoid follwoing problems.
+- https://github.com/geemus/excon/issues/106
+- https://github.com/jruby/jruby-ossl/issues/19
+
+But for all other case, it strongly reccomend to set `true` to avoid process hangin problem reported in https://github.com/uken/fluent-plugin-elasticsearch/issues/732
+
+Default value is `true`.
+
+```
+http_backend_excon_nonblock false
 ```
 
 ### compression_level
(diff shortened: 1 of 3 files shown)


ukinau:
It may be related to https://github.com/uken/fluent-plugin-elasticsearch/issues/732 . 
I explained whole story of how reach to this solution, so this may help others understand what's happening and make us on the same page
https://medium.com/uckey/why-fluentd-stopped-to-send-logs-to-elasticsearch-on-kubernetes-related-to-ssl-4ec1671b9ced

hari819:
hello ,

is this fix merged into latest release ,v1.12-debian-elasticsearch7-1?

Thanks,

data-dude:
This is a great fix.  I can't believe more people don't run into this.  If you are using Fluentd to send logs to Elastic on Kubernetes you'll run into this.

nix-power:
It looks like the very same issue happens with http out plugin. Is it suitable to apply this patch to http-out plugin code ?

PR #739, opened 2020-04-03
Use newer tls protocol versions

### LINKED ISSUE #736: SSL connect error: ssl_min_version / ssl_max_version defaults
I've started testing with latest **ubuntu 20.04** and discovered ssl connection issues. ( openssl 1.1.1d-2ubuntu6). The cause is the tls enforcement for `TLSv1` specified here https://github.com/uken/fluent-plugin-elasticsearch/blob/master/lib/fluent/plugin/elasticsearch_tls.rb (included in out_elasticsearch.rb)

~~I suggest to not setting the ssl version by default (if possible), as the openssl should decide on it's own if nothing was given.~~

This behavior seems to be introduced with this PR. https://github.com/uken/fluent-plugin-elasticsearch/pull/710

/cc  @cosmo0920 

Edit:

I've scrolled through the merge request and discovered a major change, which might cause the issue ~~(i'll verify this tomorrow)~~. (This behave the same like setting max and min version identical)
@cosmo0920 you've changed the default from passing `ssl_version` to passing `ssl_min_version` and `ssl_max_version`.

A further question: Why do you enforce setting `min` and `max` version always together. `openssl s_client -connect` could handle this as expected. 

```
openssl s_client -connect google.de:443 -min_protocol TLSv1
> New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
...
openssl s_client -connect google.de:443 -max_protocol TLSv1.2
> New, TLSv1.2, Cipher is ECDHE-RSA-CHACHA20-POLY1305
```

cosmo0920:
Fixes #736 
We should consider Elasticsearch 6.x and 7.x lines.
So, we should use newer TLS procotol version such as TLS v1.2 and TLS v1.3 by default.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -885,7 +885,7 @@ If they are *not* specified in the Elasticsearch plugin configuration, `ssl_max_
 
 In Elasticsearch plugin v4.0.8 or later with Ruby 2.5 or later environment, `ssl_max_version` should be `TLSv1_3` and `ssl_min_version` should be `TLSv1_2`.
 
-From Elasticsearch plugin v4.0.4 to v4.0.7 with RUby 2.5 or later environment, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
+From Elasticsearch plugin v4.0.4 to v4.0.7 with Ruby 2.5 or later environment, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
 
 
 ### Proxy Support
--- a/README.md
+++ b/README.md
@@ -881,7 +881,11 @@ ssl_min_version TLSv1_2
 
 Elasticsearch plugin will use TLSv1.2 as minimum ssl version and TLSv1.3 as maximum ssl version on transportation with TLS. Note that when they are used in Elastissearch plugin configuration, *`ssl_version` is not used* to set up TLS version.
 
-If they are *not* specified in the Elasticsearch plugin configuration, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
+If they are *not* specified in the Elasticsearch plugin configuration, `ssl_max_version` and `ssl_min_version` is set up with:
+
+In Elasticsearch plugin v4.0.8 or later with Ruby 2.5 or later environment, `ssl_max_version` should be `TLSv1_3` and `ssl_min_version` should be `TLSv1_2`.
+
+From Elasticsearch plugin v4.0.4 to v4.0.7 with RUby 2.5 or later environment, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
 
 
 ### Proxy Support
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin/elasticsearch_tls.rb
```diff
@@ -48,8 +48,8 @@ def set_tls_minmax_version_config(ssl_version, ssl_max_version, ssl_min_version)
       if USE_TLS_MINMAX_VERSION
         case
         when ssl_min_version.nil? && ssl_max_version.nil?
-          ssl_min_version = METHODS_MAP[ssl_version]
-          ssl_max_version = METHODS_MAP[ssl_version]
+          ssl_min_version = METHODS_MAP[:TLSv1_2]
+          ssl_max_version = METHODS_MAP[:TLSv1_3]
         when ssl_min_version && ssl_max_version.nil?
           raise Fluent::ConfigError, "When you set 'ssl_min_version', must set 'ssl_max_version' together."
```
tonobo, line 54: Why? Thats not a openssl limitation?

README.md
```diff
@@ -881,7 +881,11 @@ ssl_min_version TLSv1_2
 
 Elasticsearch plugin will use TLSv1.2 as minimum ssl version and TLSv1.3 as maximum ssl version on transportation with TLS. Note that when they are used in Elastissearch plugin configuration, *`ssl_version` is not used* to set up TLS version.
 
-If they are *not* specified in the Elasticsearch plugin configuration, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
+If they are *not* specified in the Elasticsearch plugin configuration, `ssl_max_version` and `ssl_min_version` is set up with:
+
+In Elasticsearch plugin v4.0.8 or later with Ruby 2.5 or later environment, `ssl_max_version` should be `TLSv1_3` and `ssl_min_version` should be `TLSv1_2`.
+
+From Elasticsearch plugin v4.0.4 to v4.0.7 with RUby 2.5 or later environment, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
```
tonobo, line 888:
```suggestion
From Elasticsearch plugin v4.0.4 to v4.0.7 with Ruby 2.5 or later environment, the value of `ssl_version` will be *used in `ssl_max_version` and `ssl_min_version`*.
```

lib/fluent/plugin/elasticsearch_tls.rb
cosmo0920: To prevent TLS protocol version glitch at runtime.

bastelfreak: could you elaborate what you mean with that statement? I'm not sure if I understand your statement (I'm not a native speaker).

PR #741, opened 2020-04-03
Only set Content-Encoding:gzip when body is gzip compressed

JamesJJ:
When sending a bulk index request with a gzip compressed body ( e.g. `compression_level` is not `no_compression`), then we need to set the request header `Content-Encoding: gzip` in the HTTP request.

Currently this is done globally on `client`, however that way will cause dynamic mapping template creation and possibly other requests to fail, as those requests do not have a gzip compressed request body.

This is PR is one way to resolve the problem: Only set `Content-Encoding: gzip` on the requests that have a compressed body.

This depends on https://github.com/elastic/elasticsearch-ruby/pull/820

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -525,12 +525,7 @@ EOC
           local_reload_connections = @reload_after
         end
 
-        gzip_headers = if compression
-                         {'Content-Encoding' => 'gzip'}
-                       else
-                         {}
-                       end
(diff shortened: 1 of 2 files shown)


JamesJJ:
> Could you fix failing test cases?

Hi @cosmo0920 , This depends on elastic/elasticsearch-ruby#820  ...  I'm happy to consider an alternative approach if you prefer (e.g. separate `client` for compressed and non-compressed request body)

cosmo0920:
Adding compress optional argument is one approach.

JamesJJ:
Closing this, as it has been superseded by #743

PR #799, opened 2020-09-02
Load multiple templates even if template_name and template_file given

dobesv:
This allows loading some auxiliary templates on startup even when using ILM.

If you are OK with this general approach I can adjust the other stuff (docs, tests?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ Specify index templates in form of hash. Can contain multiple templates.
 templates { "template_name_1": "path_to_template_1_file", "template_name_2": "path_to_template_2_file"}
 ```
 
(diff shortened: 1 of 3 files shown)


cosmo0920 (commented):
Looks attractive for me.
Yeah, there is no reason to be exclusive for `template_file` and `template_name`, and `templates` parameters.
Just nothing to be requested in the previous.

cosmo0920:
I updated the document for `templates` parameter: https://github.com/uken/fluent-plugin-elasticsearch/pull/799/commits/5ae1f62ff547a07624791e86b960af5ccba49ff8

PR #809, opened 2020-09-14
ElasticsearchGenID update docs for hash_type

taidv:
update docs for hash_type


cosmo0920 (approved):
Good catch! Thank you.

PR #822, opened 2020-10-14
Add a note about the pitfalls of per-date indexes used with ILM

dobesv:
Docs update


cosmo0920 (approved):
Looks reasonable.
Using curator to delete old indices is reasonable for Elasticsearch ecosystem. :+1:

Issue #797, opened 2020-09-02
Cannot use index lifecycle management with a dynamic index_name, have to use logstash_format


dobesv:
# Problem

I'm trying to use Index Lifecycle Management with index names that depend on the tag, but only a few of the indexes created by fluentd actually have lifecycle management applied.

#### Steps to replicate

elasticsearch configuration

```
<match **>
    @id elasticsearch
    @type elasticsearch
    @log_level info
    include_tag_key true
    host elasticsearch.elasticsearch
    port 9200
    default_elasticsearch_version 7
    verify_es_version_at_startup true
    logstash_format false
    enable_ilm true
    ilm_policy_id fluentd-logs
    ilm_policy {"policy":{"phases":{"hot":{"min_age":"0ms","actions":{"rollover":{"max_age":"3d","max_size":"20gb"},"set_priority":{"priority":100}}},"warm":{"actions":{"allocate":{"include":{},"exclude":{},"require":{"data":"warm"}},"set_priority":{"priority":50}}},"delete":{"min_age":"30d","actions":{"delete":{}}}}}}
    ilm_policy_overwrite false
    log_es_400_reason true
    template_overwrite true
    template_name logs
    template_file "/etc/fluent/config.d/logs-es-template.json"
    index_name "logs.${tag}"

    <buffer tag>
      @type file
      path /var/log/fluentd-buffers/kubernetes.${tag}.buffer
      flush_mode interval
      retry_max_interval 30
      flush_thread_count 2
    </buffer>
  </match>
```

`logs-es-template.json`

```
{
  "index_patterns": ["logs.*"],
  "mappings": {
    "date_detection": false,
    "dynamic_templates": [
      {
        "strings_as_keyword": {
          "mapping": { "ignore_above": 1024, "type": "keyword" },
          "match_mapping_type": "string"
        }
      },
      {
        "count_suffix_as_long": {
          "mapping": { "type": "long" },
          "match": "*Count",
          "match_mapping_type": "string"
        }
      },
      {
        "disable_unmapped_objects": {
          "mapping": { "enabled": false },
          "match_mapping_type": "object"
        }
      }
    ],
    "properties": {
      "@timestamp": { "type": "date" },
      "err": {
        "properties": {
          "code": { "ignore_above": 1024, "type": "keyword" },
          "id": { "ignore_above": 1024, "type": "keyword" },
          "message": {
            "type": "text",
            "fields": {
              "keyword": {
                "type": "keyword"
[...]

PR #823, opened 2020-10-15
Fix ILM rollover alias creation when a placeholder is used in index_name

dobesv:
This logic incorrectly assumes that if a template with the given `template_name` was previously
installed then it can safely do an early exit, even if the rollover alias was not created.

This corrects the logic so that it only does an early exit if both the rollover alias was created
AND the template was installed.

- [x] tests added - I updated existing tests to cover this scenario, that seemed to make sense
- [x] tests passing - tests for the modified code pass, but unrelated tests failed for me.  Maybe they will work in CI
- [x] backward compatible - I can't really think of a use case that this would break
- [?] feature works in `elasticsearch_dynamic` (not required but recommended) - I think it probably does, but I'm not totally sure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2006,4 +2006,7 @@ Install dev dependencies:
 $ gem install bundler
 $ bundle install
 $ bundle exec rake test
+# To just run the test you are working on:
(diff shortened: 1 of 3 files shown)


dobesv:
This is a fix for https://github.com/uken/fluent-plugin-elasticsearch/issues/797

cosmo0920 (approved):
LGTM.
Thank you for your great work! :+1:

PR #825, opened 2020-10-15
Fix ILM rollover index template pattern to apply index_separator

dobesv:
The indexes that are created when using ILM follow one of these patterns based on whether
index_date_pattern is set blank or not:

- `<index_name><index_separator><application_name>-0000001`
- `<index_name><index_separator><application_name>-<index_date_pattern>-0000001`

However, the index template always used the pattern `<index_name>-*`.  This meant if you used an
`index_separator` other than `-` the index template would not match and the ILM policy would not apply.

- [x] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920 (approved):
LGTM.
I'd overlooked `index_separator` case.
Thank you for your work! :muscle:

PR #826, opened 2020-10-15
Remove unnecessary nil check

dobesv:
The `template_name` was already checked for truthiness in the enclosing condition and cannot be `nil` on this line.


cosmo0920 (approved):
Good catch.
Thank you!

PR #838, opened 2020-11-15
fix typo

alimalek71:
Just fix a typo

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920 (approved):
Good catch. Thank you!

Issue #844, opened 2020-11-30
Retry exponential backoff does not work when using invalid/expired credentials


ventsislav-georgiev:
#### Problem
When the authentication credentials fail the plugins logs the following error:
```
2020-11-30 16:41:53 +0000 [error]: #0 unexpected error error_class=Elasticsearch::Transport::Transport::Errors::Unauthorized error="[401] {\"error\":{\"root_cause\":[{\"type\":\"security_exception\",\"reason\":\"unable to authenticate user [test] for REST request [/]\",\"header\":{\"WWW-Authenticate\":\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\"}}],\"type\":\"security_exception\",\"reason\":\"unable to authenticate user [test] for REST request [/]\",\"header\":{\"WWW-Authenticate\":\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\\\"\"}},\"status\":401}" 
2020-11-30 16:41:53 +0000 [error]: #0 /usr/lib/ruby/gems/2.5.0/gems/elasticsearch-transport-7.9.0/lib/elasticsearch/transport/transport/base.rb:218:in `__raise_transport_error'
2020-11-30 16:41:53 +0000 [error]: #0 /usr/lib/ruby/gems/2.5.0/gems/elasticsearch-transport-7.9.0/lib/elasticsearch/transport/transport/base.rb:347:in `perform_request'
2020-11-30 16:41:53 +0000 [error]: #0 /usr/lib/ruby/gems/2.5.0/gems/elasticsearch-transport-7.9.0/lib/elasticsearch/transport/transport/http/faraday.rb:37:in `perform_request'
2020-11-30 16:41:53 +0000 [error]: #0 /usr/lib/ruby/gems/2.5.0/gems/elasticsearch-transport-7.9.0/lib/elasticsearch/transport/client.rb:176:in `perform_request'
2020-11-30 16:41:53 +0000 [error]: #0 /usr/lib/ruby/gems/2.5.0/gems/elasticsearch-api-7.9.0/lib/elasticsearch/api/actions/info.rb:37:in `info'
2020-11-30 16:41:53 +0000 [error]: #0 /usr/lib/ruby/gems/2.5.0/gems/fluent-plugin-elasticsearch-4.2.2/lib/fluent/plugin/out_elasticsearch.rb:482:in `detect_es_major_version'
[...]

PR #846, opened 2020-12-01
feat: retry on transport errors

ventsislav-georgiev:
Related to: https://github.com/uken/fluent-plugin-elasticsearch/issues/844

(check all that apply)
- [x] tests added
- [x] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/elasticsearch_index_template.rb
+++ b/lib/fluent/plugin/elasticsearch_index_template.rb
@@ -32,10 +32,10 @@ module Fluent::ElasticsearchIndexTemplate
     return false
   end
 
-  def retry_operate(max_retries, fail_on_retry_exceed = true)
+  def retry_operate(max_retries, fail_on_retry_exceed = true, catch_trasport_exceptions = true)
     return unless block_given?
     retries = 0
-    transport_errors = Elasticsearch::Transport::Transport::Errors.constants.map{ |c| Elasticsearch::Transport::Transport::Errors.const_get c }
+    transport_errors = Elasticsearch::Transport::Transport::Errors.constants.map{ |c| Elasticsearch::Transport::Transport::Errors.const_get c } if catch_trasport_exceptions
     begin
       yield
     rescue *client.transport.host_unreachable_exceptions, *transport_errors, Timeout::Error => e
--- a/lib/fluent/plugin/elasticsearch_index_template.rb
+++ b/lib/fluent/plugin/elasticsearch_index_template.rb
@@ -35,9 +35,10 @@ module Fluent::ElasticsearchIndexTemplate
   def retry_operate(max_retries, fail_on_retry_exceed = true)
     return unless block_given?
     retries = 0
+    transport_errors = Elasticsearch::Transport::Transport::Errors.constants().map{ |c| Elasticsearch::Transport::Transport::Errors.const_get c }
     begin
       yield
-    rescue *client.transport.host_unreachable_exceptions, Timeout::Error => e
+    rescue *client.transport.host_unreachable_exceptions, *transport_errors, Timeout::Error => e
       @_es = nil
       @_es_info = nil
       if retries < max_retries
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/elasticsearch_index_template.rb
```diff
@@ -35,9 +35,10 @@ def template_exists?(name, host = nil)
   def retry_operate(max_retries, fail_on_retry_exceed = true)
     return unless block_given?
     retries = 0
+    transport_errors = Elasticsearch::Transport::Transport::Errors.constants().map{ |c| Elasticsearch::Transport::Transport::Errors.const_get c }
```
cosmo0920, line 38: I think that all errors for transporting is awful handling.
For example, `BadRequest`, or `MethodNotAllowed`, or `UnsupportedMediaType` requests shouldn't repeat.

cosmo0920, line 38: `constants()` should be `constants` (without empty parentheses).

test/plugin/test_out_elasticsearch.rb
```diff
@@ -3052,6 +3052,42 @@ def test_template_retry_install_fails(data)
     assert_equal(4, connection_resets)
   end
 
+  transport_errors_handled_separately = [Elasticsearch::Transport::Transport::Errors::NotFound]
+  transport_errors = Elasticsearch::Transport::Transport::Errors.constants().map { |err| [err, Elasticsearch::Transport::Transport::Errors.const_get(err)]  }
```
cosmo0920, line 3056: constants() should be constants (without empty parentheses).

cosmo0920 (changes requested):

lib/fluent/plugin/elasticsearch_index_template.rb
ventsislav-georgiev: @cosmo0920 
Let me explain the issue we are having. We are exposing a fluentd as a service, users can setup their own output destination to receive logs from our backend. This is working inside Kubernetes with auto scaling based on CPU. The issue is that if they do not set it correctly this crash and repeat every couple of seconds is hitting 100% CPU usage and the HorizontalPodAutoscaler starts scaling the invalid services. The error returned may as well be BadRequest from the destination they have set up.

We either need a hard crash in these cases so that the pod can simple go into a crashloop or this retry, so that their CPU usage is not 100%.

ventsislav-georgiev: I looked into Fluentd Output buffer code and saw that the issue comes when an error is raised in the `configure` or `start` phase and the buffer retry logic is only happening with errors during the `write(chunk)` phase. Maybe we should fork Fluentd and apply a hard crash on errors in the first two phases.

The Splunk plugin amazed me by having intentionally consuming a chunk when any of the 4xx http responses is returned: https://github.com/splunk/fluent-plugin-splunk-hec/blob/develop/lib/fluent/plugin/out_splunk_hec.rb#L308

We are using it with Kafka and consuming a chunk will be a data loss for the specific user since their consumer offset is getting committed.

cosmo0920: I see. Pokemon exception for `transport_errors` are reasonable for me in this case.
Thank you for the explanation.

cosmo0920 (commented):
Could you fix trailing parentheses on `constants()`?
Other concerns are just needless fear.
Good work! :+1:

ventsislav-georgiev:
@cosmo0920 Some part of these changes breaks the tests. They successfully run locally. Any idea what is the issue?

cosmo0920:
Thanks for your work!
I'd checked the codebase. Then I found that some testcases depend on previous behavior on `#retry_operate`.
I'd introduced `catch_transport_exception_on_retry` parameter to restore previous behavior. (Mainly this parameter is for unit testing.)

cosmo0920 (approved):

PR #870, opened 2021-02-25
Use elasticsearch/api instead of elasticsearch/xpack

kenhys:
Since elasticseach-ruby 7.11.0 indices.put_index_template is moved to elasticsearch/api.

ref. https://github.com/elastic/elasticsearch-ruby/blob/master/CHANGELOG.md#7110


cosmo0920 (approved):
Oh! I didn't notice about it.
Thanks for your work. :+1:

Issue #871, opened 2021-03-01
Data stream templates and ILM


snemetz:
**Is your feature request related to a problem? Please describe.**
I have component templates, index templates, and ILM policies on ES.
But during fluentd startup, default index templates and ILM policies are written for data streams. These overwrite any existing templates and ILM. These are written whether data stream already exists or not. This prevents the use of component templates and setups on ES. This is also a different behavior than what is done for indices

**Describe the solution you'd like**
1 of 2 options.

1. Do not write templates and ILM unless specified. This is the behavior used for indices. Since data streams require templates, fail if data stream write/create fails
2.  Only write default template and ILM if data stream write and create fail

**Describe alternatives you've considered**

Since current behavior overwrites on startup. Fluentd can not be in an autoscaling group.
But for static servers, I will be testing the following:

1. Create templates and ILM
2. Create data stream
3. Write something to data stream
4. start fluentd
5. upload templates and ILM again

This is a bit of a pain. But It might work temporarily

wbrefvem:
> Fluentd can not be used in an autoscaling group

Or, to put it another way, this plugin can't be used in Kubernetes without significant workarounds. If you're running a fluentd daemonset to scrape k8s pod logs, any time a fluentd pod dies -- or in fact any time the fluentd container restarts -- the ILP and index template associated with the data stream will be overwritten. There's currently no way to decouple these.

Looks like there's a simple fix -- just don't update the ILM policy or the index template if the data stream already exists.

cosmo0920:
At master has temporary fix but we should still open this issue for _final fixes/improvements_.

srook:
Are there any plans on implementing first suggested solution? Parameter to skip creation of template/ILM?

User should be able to decide if templates/ILMs should be managed through Fluent or Elastic, and not be forced to manage them through Fluent.

PR #872, opened 2021-03-15
Fix overwriting of index template and index lifecycle policy on existing data streams

wbrefvem:
Addresses https://github.com/uken/fluent-plugin-elasticsearch/issues/871

Currently, the `elasticsearch_data_stream` type creates a data stream, along with associated index template and index lifecycle policy. The `create_data_stream` method checks whether the data stream exists before creating it. We should do the same before overwriting the index template and index lifecycle policy.

This update simply runs `data_stream_exist?` at the top of `create_ilm_policy` and `create_index_template`, as is currently done in `create_data_stream`. The current approach leaves room for adding support for passing in an ILM policy and index template, but that's currently not supported. These calls to `data_stream_exist?` will need to be removed if such support is added (which is ultimately a more robust solution long-term).

With this change, users will be able to update the ILM policy and the index template after creation, and those updates won't be clobbered every time a fluentd process starts up (e.g., if a fluentd pod on Kubernetes dies and is rescheduled). 

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch_data_stream.rb
+++ b/lib/fluent/plugin/out_elasticsearch_data_stream.rb
@@ -67,6 +67,7 @@ module Fluent::Plugin
     end
 
     def create_ilm_policy(name)
+      return if data_stream_exist?(name)
       params = {
         policy_id: "#{name}_policy",
         body: File.read(File.join(File.dirname(__FILE__), "default-ilm-policy.json"))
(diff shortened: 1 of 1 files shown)


wbrefvem:
@kenhys Do you have any thoughts on this?

kenhys:
It is reasonable about this approach.
Could you add test case?

cosmo0920 (approved):
I decide merging this PR due to sending with master branch.... :cry:
Sending PR via master branch does not permit to additional changes by PR sent repository developer.
We'll try to add testcases for this later.

PR #880, opened 2021-04-15
Fix use_legacy_template documentation

haines:
The `use_legacy_template` documentation in the README currently says that it should be set to `false` for Elasticsearch <= 7.7. I think this is backwards - the setting should be `true` (the default) if using the legacy templates and only `false` if using the new composable templates with Elasticsearch >= 7.8.

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -1353,9 +1353,9 @@ Default value is `nil`.
 
 Use legacy template or not.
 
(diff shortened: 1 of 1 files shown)


cosmo0920 (approved):
Ah, that's true.
Good catch. Thank you! :+1:

PR #900, opened 2021-06-22
Drop json_parse_exception messages for bulk failures

jcantrill:
When Elasticsearch rejects a record because it is unable to parse the content(i.e due to an encoding issue), the record blocks resubmission of the remaining records since the specific record can never be parsed by the server.  

This PR:
* Add functionality to drop specific status:500 records which are not recoverable

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1951605

(check all that apply)
- [x]  tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/lib/fluent/plugin/elasticsearch_error_handler.rb
+++ b/lib/fluent/plugin/elasticsearch_error_handler.rb
@@ -23,6 +23,10 @@ class Fluent::Plugin::ElasticsearchErrorHandler
     unrecoverable_error_types.include?(type)
   end
 
+  def unrecoverable_record_error?(type)
+    ['json_parse_exception'].include?(type)
+  end
+
   def log_es_400_reason(&block)
     if @plugin.log_es_400_reason
       block.call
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/elasticsearch_error_handler.rb
```diff
@@ -23,6 +23,10 @@ def unrecoverable_error?(type)
     unrecoverable_error_types.include?(type)
   end
 
+  def unrecoverable_record_error?(type)
+    ['json_parse_exception'].include?(type)
```
cosmo0920, line 27: It's pretty good for adding unrecoverable record error concept, but I think that `es_rejected_execution_exception` is also unrecoverable exception.

What do you think about this?
Keep as-is for now is also OK for me.

cosmo0920:
Your PR concept is pretty good for me.
I added a comment for small concerns.

lib/fluent/plugin/elasticsearch_error_handler.rb
jcantrill: I defer to you here as I do not know.  If this is an exception one might expect from trying to push an individual record and it is not possible to recover or not possible to successfully resubmit, I agree we can add this to the list

jcantrill:
> Your PR concept is pretty good for me.
> I added a comment for small concerns.

Please advise if you wish for me to make changes before merging

jcantrill:
@cosmo0920 ... ping.. please advise

cosmo0920:
Sorry to be late merging. I decided this PR merge as-is.

Issue #805, opened 2020-09-10
Got "buffer flush took longer time than slow_flush_log_threshold" error


chikinchoi:
(check apply)
- [ ] read [the contribution guideline](https://github.com/uken/fluent-plugin-elasticsearch/blob/master/CONTRIBUTING.md)

#### Problem
 Hi Team,

I got at lot of `buffer flush took longer time than slow_flush_log_threshold: elapsed_time=60.16122445899964 slow_flush_log_threshold=20.0 plugin_id="firelens_es` error in my fluentd which is connecting to Elasticsearch .
I saw that there is a document mentioned the reason for this error is because  Elasticsearch is exhausted [1]. However, I checked all of the Elasticsearch clusters' CPU usage are low and healthy.

I have asked in Fluentd group and @repeatedly said that I should check the Fluentd CPU usage. Therefore, I found that the Fluentd service CPU usage is very high (almost 100%) when slow_flush error appears. I have increased the Fluentd CPU size and the error seems reduced but still can find it. I have tried to increase the Fluentd CPU size from 1024 to 2048. Although the number of slow_flush errors is decreased, I can still see it sometimes.
In conclusion, I think that the slow_flush error cause the Fluentd CPU usage increase. Therefore, I would like to know what is the cause of this error and how to fix it. Also, I don't understand why I should check the Fluentd CPU usage instead of the Elasticsearch CPU usage. 
Please advise. Thank you very much!!

#### Steps to replicate

cannot reproduce because I don't know the reason for this error.
Below is my Fluentd config:

```
<system>
  @log_level debug
  workers 4
  root_dir /var/log/fluent/
</system>

<source>
  @type  forward
  @id    input1
  @log_level debug
  port  24224
  bind 0.0.0.0
</source>

# Used for docker health check
<source>
  @type http
  port 8888
  bind 0.0.0.0
</source>

# records sent for health checking won't be forwarded anywhere
<match health**>
  @type null
</match>

# Prometheus Configuration
# count number of incoming records per tag
<filter **>
  @type prometheus
  <metric>
    name fluentd_input_status_num_records_total
    type counter
    desc The total number of incoming records
    <labels>
      tag ${tag}
      hostname ${hostname}
    </labels>
  </metric>
</filter>

<filter **firelens**>
  @type concat
  key log
  multiline_start_regexp '^\{\\"@timestamp'
  multiline_end_regexp '/\}/'
  separator ""
  flush_interval 1
[...]

PR #914, opened 2021-08-19
Correct default target bytes value

cosmo0920:
Signed-off-by: Hiroshi Hatake <hatake@calyptia.com>

Related to #805.

Default value of bulk_message_request_threshold should be `-1` not 20MB (=20 * 1024 * 1024)
This causes performance degradation for ES plugin.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -72,7 +72,7 @@ module Fluent::Plugin
     DEFAULT_TYPE_NAME_ES_7x = "_doc".freeze
     DEFAULT_TYPE_NAME = "fluentd".freeze
     DEFAULT_RELOAD_AFTER = -1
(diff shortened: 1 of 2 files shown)


kenhys (commented):
Is unlimited bulk message always reasonable?
I have a concern about the value may be too small in the previous 20MB threshold.
Instead, a more proper default value may exist. 
(even though I do not have confidence)

If unlimited is best, It is ok for me.
(In the other hand, it is reasonable that  If something wrong, bulk_message_request_threshold should be customized)

kenhys:
Oops, there are mismatch between documentation and implementation.
https://github.com/uken/fluent-plugin-elasticsearch#bulk_message_request_threshold

kenhys (approved):
LGTM.

cosmo0920:
Yeah, documentation and implementation mismatch existed until now....

Issue #904, opened 2021-06-28
"enable_ilm = false" when @type elasticsearch_data_stream is used


JPIFR:
It's written in the doc: "**Configuration - Elasticsearch Output Data Stream**": 

**When @type elasticsearch_data_stream is used, ILM default policy is set to the specified data stream. Then, the matching index template is also created automatically.**

We would like to disable all automatic creation of Fluentd elastcisearch plugin because we use an index template which creates and manage already everything.  

Are you planning to develop such a feature?

Thanks for your help!

cosmo0920:
No. We don't plan to support such feature.

expteam-interactiv:
I have the same need using a custom index template/ILM with data streams

PR #920, opened 2021-09-23
Add ILM and template parameters for data streams

fiscafusca:
I added two optional parameters (`data_stream_ilm_name` and  
`data_stream_template_name`) that can be used to specify the name  
of already existing ILM policy and template. If not present in the  
configuration, the default policy and template will be created,  
as it was before.  
The relative test file was edited to work with the new parameters.

When creating a new data stream, one may want to specify an existing  
ILM policy and/or index template to be applied to the data stream  
(see [issue #904](https://github.com/uken/fluent-plugin-elasticsearch/issues/904)). Having pre-made templates with a higher priority is  
just a patch that does not solve the root of the problem.  
For example, in our application we need to generate multiple data  
streams from fluentd. Without this feature, the plugin ends up  
generating loads of identical default ILM policies and templates that  
remain unused and must be deleted manually.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -21,4 +21,4 @@ require 'fluent/test'
 require 'minitest/pride'
 
 require 'webmock/test_unit'
-WebMock.enable_net_connect!
+WebMock.disable_net_connect!
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -21,4 +21,4 @@ require 'fluent/test'
 require 'minitest/pride'
 
 require 'webmock/test_unit'
-WebMock.disable_net_connect!
+WebMock.enable_net_connect!
--- a/README.md
+++ b/README.md
@@ -1521,13 +1521,25 @@ You can enable this feature by specifying `@type elasticsearch_data_stream`.
 data_stream_name test
 ```
 
-When `@type elasticsearch_data_stream` is used, ILM default policy is set to the specified data stream.
+When `@type elasticsearch_data_stream` is used, unless specified with `data_stream_ilm_name` and `data_stream_template_name`, ILM default policy is set to the specified data stream.
 Then, the matching index template is also created automatically.
 
 ### data_stream_name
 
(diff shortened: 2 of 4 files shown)


cosmo0920 (changes requested):
Unit test cases are failed. Could you check them?

fiscafusca:
Sorry, I was out of office yesterday. I just checked, it should work now

test/helper.rb
```diff
@@ -21,4 +21,4 @@
 require 'minitest/pride'
 
 require 'webmock/test_unit'
-WebMock.disable_net_connect!
+WebMock.enable_net_connect!
```
cosmo0920, line 24: You shouldn't do this.
Please revert this line.

cosmo0920 (changes requested):

test/helper.rb
fiscafusca: Just reverted. Sorry, I'm kinda new with these tools. The tests were doing fine with this directive locally, I didn't know I shouldn't change that line. 
I updated the PR message, as the work in its current state is clearly not passing the tests. I will look for errors and shortcomings in my code in the upcoming days.

fiscafusca:
Fixed a couple of things, the tests do fine now on my machine without changing anything aside the data stream file and its relative test file. Hopefully nothing else should be missing

cosmo0920 (approved):
Looks good. I'll fix up tiny code style breaks.

cosmo0920:
Hi @fiscafusca , I sent a PR for following up your work.
Ordinary ES plugin users doesn't want to read its implementation.
Invalid Data Stream parameter related error messages should be reported with actual parameter names.

fiscafusca:
Hi @cosmo0920, just checked your PR! You're right, I definitely missed this. Thank you for following up

PR #933, opened 2021-12-18
allow specifying custom ILM policies for data streams

BitPatty:
This PR allows to specify a custom ILM policy when using data streams (and to overwrite an existing policy if necessary).

I am not sure why the current implementation skips the ILM policy creation if the data stream or template exist, but I didn't touch this check for backwards compatibility.
https://github.com/uken/fluent-plugin-elasticsearch/blob/4a65040cdbbc5d6bf05e73f5a29dcbb35f653521/lib/fluent/plugin/out_elasticsearch_data_stream.rb#L80

(note: I usually don't write ruby code, so if you have suggestions for improvements let me know)

(check all that apply)
- [x] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -1521,7 +1521,7 @@ You can enable this feature by specifying `@type elasticsearch_data_stream`.
 data_stream_name test
 ```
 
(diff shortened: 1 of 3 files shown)


cosmo0920 (approved):
Looks good. Thanks for the contribution.

PR #960, opened 2022-04-04
Ensure use_record_as_seed for same records

sihyeonn:
`use_record_as_seed` cannot ensure same result for same record but different order of keys when `use_entire_record`.
To ensure `use_record_as_seed` with `use_entire_record`, I changed code as using sorted keys to make seed.

Signed-off-by: Sihyeon Jang <uneedsihyeon@gmail.com>

DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [x] tests passing
- [ ] backward compatible


cosmo0920 (approved):
Seems reasonable. Thanks for the contribution! :+1:

PR #985, opened 2022-09-07
Update fluent-plugin-elasticsearch.gemspec

josevromo:
DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible


cosmo0920 (changes requested):
Could you align runtime as development dependency.

PR #986, opened 2022-09-14
Increase errors metric on error response in data stream

mkmet:
If ES responds with an error (e.g., because of a type issue), this isn't visible in any metric (only an error log is generated). Because of this, monitoring solutions based on metrics scraping (like Prometheus) can't detect this issue. This patch increases num_errors metric if there are errors in response from ES in the data stream output.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible


cosmo0920 (approved):
Good catch, Thanks!

PR #1012, opened 2023-04-03
build(deps): Unpin `faraday` from v1, upgrade to v2

raytung:
DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.7', '3.0', '3.1', '3.2' ]
+        ruby: [ '3.0', '3.1', '3.2' ]
         os:
           - ubuntu-latest
     name: Ruby ${{ matrix.ruby }} unit testing on ${{ matrix.os }}
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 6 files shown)


raytung:
Thanks @cosmo0920 for helping out on this issue - I believe this is a confluent of issues, and there's a bug with Faraday and its adaptor in Ruby 2.7 (which I've jotted down in the linked issue). I don't think we could drop Ruby 2.7 support at the moment since the latest official Fluentd docker image is running on Ruby 2.7. Interested to hear your thoughts on this one.

cosmo0920:
From the official Fluentd docker image, we already moved to use Ruby 3.1 instead of 2.7.
The left things are still using docker official image that is using Ruby 2.7:
https://github.com/docker-library/official-images/blob/381271cfcf579d69b2a1410594e2926f865d38df/library/fluentd#L6
which is based on https://github.com/fluent/fluentd-docker-image/commit/a9e4cee765c7aaf7876d6fe3282aa565dbcdc2a4.

raytung:
@cosmo0920 ah yes of course, I must be blind. Sounds like this fix should work (if all the tests pass). Thanks for your help!

cosmo0920 (approved):
A big :+1:

cosmo0920:
@kenhys @ashie Could you release the new version of this plugin if needed?

ashie:
I see, thanks!

cosmo0920:
FYI: Docker library image of Fluentd is now updated to use Fluentd v1.16.0 and Ruby 2.7 images are completely gone!

PR #1044, opened 2024-01-05
Avoid long worrying log line at info level

sparrowt:
Currently there is an info level log line when checking if a datastream exists, but it includes the entire error even though it is caught and expected.

This results in long worrying looking log messages which are in fact not a problem (once you notice that they are info level) e.g. like this:
> [info]: #0 Specified data stream does not exist. Will be created: <[404] {"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index [my.datastream.name]","resource.type":"index_or_alias","resource.id":"my.datastream.name","index_uuid":"_na_","index":"my.datastream.name"}],"type":"index_not_found_exception","reason":"no such index [my.datastream.name]","resource.type":"index_or_alias","resource.id":"my.datastream.name","index_uuid":"_na_","index":"my.datastream.name"},"status":404}>

We know what the error is going to be given it is being caught explicitly, so instead simplify and just log the name of the datastream which looks much nicer like this instead:
> [info]: #0 Specified data stream does not exist. Will be created: <my.datastream.name>

I don't think any of the below checklist are relevant but please let me know if you think they are.
Thanks!

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [ ] backward compatible

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch_data_stream.rb
+++ b/lib/fluent/plugin/out_elasticsearch_data_stream.rb
@@ -142,7 +142,7 @@ module Fluent::Plugin
         response = client(host).indices.get_data_stream(params)
         return (not response.is_a?(TRANSPORT_CLASS::Transport::Errors::NotFound))
       rescue TRANSPORT_CLASS::Transport::Errors::NotFound => e
-        log.info "Specified data stream does not exist. Will be created: <#{e}>"
+        log.info "Specified data stream does not exist. Will be created: <#{datastream_name}>"
         return false
       end
     end


lib/fluent/plugin/out_elasticsearch_data_stream.rb
```diff
@@ -142,7 +142,7 @@ def data_stream_exist?(datastream_name, host = nil)
         response = client(host).indices.get_data_stream(params)
         return (not response.is_a?(TRANSPORT_CLASS::Transport::Errors::NotFound))
       rescue TRANSPORT_CLASS::Transport::Errors::NotFound => e
-        log.info "Specified data stream does not exist. Will be created: <#{e}>"
+        log.info "Specified data stream does not exist. Will be created: <#{datastream_name}>"
```
cosmo0920, line 145: How about decrease log level for entire log and add the proposed line?
This should be more reasonable to suppress log warnings. What do you think?

cosmo0920: Noted. Let's change the log message. No need to further action.

cosmo0920 (approved):

lib/fluent/plugin/out_elasticsearch_data_stream.rb
sparrowt: Ok thanks! Is there anything else I need to do before you can merge this?

cosmo0920: There is no action for you. I'm just late for merging this. Thank you for your work. Good catch. :smile:

sparrowt: No worries, thank you!

Issue #1061, opened 2025-04-26
Fluentd to Elasticsearch error: Invalid media-type value on headers [Content-Type, Accept]


Srijitha-code:
I am encountering an error when Fluentd tries to ship logs to an Elasticsearch cluster. The error is:
`2025-04-26 14:04:47 +0000 [warn]: #0 fluent/log.rb:383:warn: failed to flush the buffer. retry_times=3 next_retry_time=2025-04-26 14:04:55 +0000 chunk="633aeee3e06bda3cf586d204d14f472c" error_class=Fluent::Plugin::ElasticsearchOutput::RecoverableRequestFailure error="could not push logs to Elasticsearch cluster ({:host=>\"quickstart-v2-es-http\", :port=>9200, :scheme=>\"https\", :user=>\"elastic\", :password=>\"obfuscated\"}): [400] {\"error\":{\"root_cause\":[{\"type\":\"media_type_header_exception\",\"reason\":\"Invalid media-type value on headers [Content-Type, Accept]\"}],\"type\":\"media_type_header_exception\",\"reason\":\"Invalid media-type value on headers [Content-Type, Accept]\",\"caused_by\":{\"type\":\"status_exception\",\"reason\":\"A compatible version is required on both Content-Type and Accept headers if either one has requested a compatible version. Accept=null, Content-Type=application/vnd.elasticsearch+x-ndjson; compatible-with=9\"}},\"status\":400}"
`

Environment Details:

Fluentd Version: 1.18.0

Fluentd Gems:

fluent-plugin-elasticsearch version 5.4.3

elastic-transport version 8.4.0

elasticsearch version 9.0.2

elasticsearch-api version 9.0.2

Fluentd Deployment: Kubernetes Deployment

Elasticsearch Deployment: ECK (Elastic Cloud on Kubernetes) Operator 3.0

Elasticsearch Version: 9.0

OS Details: Debian GNU/Linux 12 (Bookworm) [ARM64]

Fluentd Config for Elasticsearch Output Plugin

```aconf
<match local.test>
  @type elasticsearch
  host ******
  port 9200
  scheme https
  user elastic
  password *****
  verify_es_version_at_startup true
  default_elasticsearch_version 9
  ca_file /fluentd/etc/ca.crt
  index_name test-logs
  content_type application/x-ndjson
  logstash_format false
  include_timestamp true
  time_key @timestamp
  <buffer>
    @type file
    path /fluentd/buffer/local_test
[...]

PR #1062, opened 2025-04-30
Fix Elasticsearch‐gem version constraints to avoid HTTP header errors

yugeeklab:
DESCRIPTION HERE

(check all that apply)
- [ ] tests added
- [ ] tests passing
- [x] backward compatible

After upgrading the elasticsearch Ruby gem to v9.0.1, Fluentd’s fluent-plugin-elasticsearch started raising:

Invalid media-type value on headers [Content-Type, Accept]

See: [uken/fluent-plugin-elasticsearch#1061](https://github.com/uken/fluent-plugin-elasticsearch/issues/1061)
https://github.com/elastic/elasticsearch-ruby/pull/2660

To restore compatibility:

For users on Elasticsearch-gem 8.x, continue shipping plugin v5.x

For users on Elasticsearch-gem 9.x, bump plugin to v6.x

This change pins the appropriate plugin version to each major gem series, preventing the header‐validation error.

### DIFF
--- a/fluent-plugin-elasticsearch.gemspec
+++ b/fluent-plugin-elasticsearch.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'faraday', '>= 2.0.0'
   s.add_runtime_dependency 'faraday-excon', '>= 2.0.0'
   s.add_runtime_dependency 'excon', '>= 0'
-  s.add_runtime_dependency 'elasticsearch'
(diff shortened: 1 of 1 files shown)


yugeeklab:
@cosmo0920

cosmo0920 (changes requested):
We don't want to lock versions for dependent gems. This could be achieved by Gemfile.
https://docs.fluentd.org/deployment/plugin-management#gemfile-option

PR #1064, opened 2025-05-01
Adjust Content-Type and Accept for sending logs into ES9

cosmo0920:
Trying to process #1061.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible

Tested with the following configuration:

```aconf
  <source>
    @type dummy
    tag "test.local"
  </source>
  <match test.local>
    @type elasticsearch
    host "localhost"
    port 9200
    scheme https
    user "elastic"
    password xxxxxx
    verify_es_version_at_startup true
    default_elasticsearch_version 9
    ca_file "http_ca.crt"
    index_name "test-logs"
    logstash_format false
    include_timestamp true
    time_key "@timestamp"
  </match>
```

### DIFF
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -147,7 +147,7 @@ EOC
     config_param :sniffer_class_name, :string, :default => nil
     config_param :selector_class_name, :string, :default => nil
     config_param :reload_after, :integer, :default => DEFAULT_RELOAD_AFTER
-    config_param :content_type, :enum, list: [:"application/json", :"application/x-ndjson", :"#{ES9_CONTENT_TYPE}"], :default => :"application/json",
+    config_param :content_type, :enum, list: [:"application/json", :"application/x-ndjson"], :default => :"application/json",
                  :deprecated => <<EOC
 elasticsearch gem v6.0.2 starts to use correct Content-Type. Please upgrade elasticserach gem and stop to use this option.
 see: https://github.com/elastic/elasticsearch-ruby/pull/514
--- a/lib/fluent/plugin/out_elasticsearch.rb
+++ b/lib/fluent/plugin/out_elasticsearch.rb
@@ -73,6 +73,7 @@ module Fluent::Plugin
     DEFAULT_RELOAD_AFTER = -1
     DEFAULT_TARGET_BULK_BYTES = -1
     DEFAULT_POLICY_ID = "logstash-policy"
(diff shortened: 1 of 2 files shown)


cosmo0920:
Could you double-check this patch, @daipom or @kenhys?

daipom:
Thanks so much for this improvement!
 I can't check very carefully, but I will as far as possible.

daipom:
Sorry, could you please tell me if my understanding is correct?
Do we need to specify `application/vnd.elasticsearch+x-ndjson; compatible-with=9`both to the following header?

* `Accept`
* `Content-Type`

cosmo0920:
> Sorry, could you please tell me if my understanding is correct? Do we need to specify `application/vnd.elasticsearch+x-ndjson; compatible-with=9`both to the following header to send data to ES9?
> 
>     * `Accept`
> 
>     * `Content-Type`

Yes, they need to configure and use within the clients.

daipom:
Thanks!

lib/fluent/plugin/out_elasticsearch.rb
```diff
@@ -146,7 +147,7 @@ def initialize(retry_stream)
     config_param :sniffer_class_name, :string, :default => nil
     config_param :selector_class_name, :string, :default => nil
     config_param :reload_after, :integer, :default => DEFAULT_RELOAD_AFTER
-    config_param :content_type, :enum, list: [:"application/json", :"application/x-ndjson"], :default => :"application/json",
+    config_param :content_type, :enum, list: [:"application/json", :"application/x-ndjson", :"#{ES9_CONTENT_TYPE}"], :default => :"application/json",
```
daipom, line 150: Could you tell me why `ES9_CONTENT_TYPE` is added to this deprecated param?

It looks to me that a user does not need to set this param since the `Content-Type` and `Accept` headers are automatically fixed when the ES version is greater than or equal to `9`.
Setting this value explicitly to `ES9_CONTENT_TYPE` would cause `@accept` to be `nil`.
Is this intentional?

cosmo0920: Oh, that's only left things for tryal-and-error investigation process.
I'll remove it from there.

daipom: I see!

daipom (approved):
LGTM! Thanks!

cosmo0920:
Hey @kenhys,
Once you finalized to bump up 5.4.4, could you double-check this PR and release this as v6.0.0 if you'd OK.

PR #1084, opened 2026-07-03
Allow duplicate keys in JSON parser

Watson1978:
The json gem emits a deprecation warning for duplicate keys in JSON objects and will raise `JSON::ParserError` in json 3.0 unless `allow_duplicate_key: true` is specified.

The plugin has historically accepted duplicate keys silently (last value wins).
To preserve this behavior uniformly regardless of the json gem version, this passes `allow_duplicate_key: true` to `JSON.parse`.

Note that `allow_duplicate_key` is silently ignored by json < 2.13, so this is safe with older json gem versions.

(check all that apply)
- [ ] tests added
- [x] tests passing
- [x] backward compatible


Watson1978:
@kenhys Can you review this PR? Thanks.

kenhys (approved):

